PR #52, opened 2011-03-28
Tag and group feature

medihack:
Sometimes it's necessary to group several audits somehow. This is especially true when auditing associations. Think of a document that has_many properties (set to autosave => true). When some properties of a document are changed and the document is saved each property gets an own audit entry in the audit table. One may group such a saving by using the created_at time, but that is not very exact. When there are many properties saved with the document then the created_at time may differ in one or more seconds. I also didn't see an easy way to set multiple times the same comment at once.
That is why I added an as_group and audit_with method that sets a provided tag to all audits and the same comment.
A tag could also be set to the audited model by using audit_tag. Note that a tag is not the same as comment. A tag is more for internal use and comment is normally set by the user.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -77,9 +77,9 @@ To record a user in the audits outside of a web request, you can use +as_user+:
       # Perform changes on audited models
     end
 
-It is possible to group several audits and give them the same +tag+ and/or +comment+. You can use +as_group+ for that:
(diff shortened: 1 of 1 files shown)


ryanwjackson:
Is there a reason this wasn't merged?

tbrisker:
This PR has been open for many years now with no activity and the main code has significantly diverged from this branch. @medihack, do you think it's still needed and would you be interested in returning to this or should we close it?

tbrisker:

PR #58, opened 2011-05-03
fixing case when the variable "controller" is not defined

punitrathore:
All my rSpec tests were failing with this error -

  You have a nil object when you didn't expect it!
  The error occurred while evaluating nil.session

Testing for the variable "controller" fixed it. 

FYI, I'm on Rails 2.3.3, and I'm using restful-authentication.


kennethkalmer:
@punitrathore thanks for the contribution! Please accept my apologies on behalf of the team for letting this age out so badly. If this is still an issue that needs to be fixed with the latest versions, please open up a new issue.

PR #60, opened 2011-05-22
1.1 stable

johndouthat:
in 1.1, if the association name is not derivative of the ActiveRecord class name, then it can't find the ActiveRecord class.

for example:

```ruby
class Foo < ActiveRecord::Base
  acts_as_audited
end

class FooRelationship < ActiveRecord::Base
  belongs_to :parent_foo, :class_name => "Foo"
  belongs_to :child_foo, :class_name => "Foo"
  acts_as_audited :parent => :parent_foo # this will throw an exception
end
```


kennethkalmer:
@johndouthat thanks for the contribution! Please accept my apologies on behalf of the team for letting this age out so badly. If this is still an issue that needs to be fixed with the latest versions, please open up a new issue.

PR #67, opened 2011-08-29
Mass assignment error

dels:
Without this fix on save I got this mass assignment error: Mass assignment error: ["action", "audited_changes", "comment"]

This becomes fixed by adding the line "attr_accessible :action, :audited_changes, :comment" to the audit.rb.


kennethkalmer:
Please continue the discussion in #186 if this is still relevant to you. Like I mentioned there, I'm worried that having `action` & `audited_changes` accessible for mass-assignment allows malicious users to easily tamper with the audits and thus render them useless...

PR #79, opened 2011-12-12
If a record is destroyed before it ever gets saved then it should be ignored.

scnaegle:
...nored.


kennethkalmer:
Seems this was fixed in 1167bce8 over a year ago by @rpond-pa. Apologies for letting this hang around so long

PR #78, opened 2011-12-01
lazy non_audited_columns method

lest:
Fetch table column names only when they are needed for first time to avoid load errors when model table doen't exists on model load.

PR #99, opened 2012-05-08
Introduce non_auditable_columns lazy method.

hron:
This is adopted version of pull request #78. I do not understand why it was not applied before...

Let me clarify what use case it resolves. When an user of audited creates a model and pass :only option, e.g.

```ruby
class User < ActiveRecord::Base
  audited :only => [:administrator]
end
```

then `rake db:migrate` will constantly fail because Audited will call `column_names` on a model without table.

I am sure this is a typical use case and every user of audited will be frustrated when it will encounter such an error. ;)

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -49,17 +49,11 @@ module Audited
 
         options = { :protect => accessible_attributes.blank? }.merge(options)
 
-        class_attribute :non_audited_columns,   :instance_writer => false
+        class_attribute :audit_options,         :instance_writer => false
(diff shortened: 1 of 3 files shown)


piya23300:
+1
PR #78 #99 work!

RohitRox:
Why this fix isn't applied yet? Can we please have this merged?

ankane:
Was just about to open a pull request with this exact feature. https://github.com/ankane/audited/commit/6491337edceb3e481a26fcd33df2f664abc38d95

:+1:

tbugai:
This has been implemented in newer versions of the gem.

Issue #101, opened 2012-05-11
Audit doesn't get pulled in automatically with rc1


ralfthewise:
I had to create an initializer with:

include Audited::Adapters::ActiveRecord

in it in order to be able to do things like:

Audit.order("created_at desc").limit(10)

it doesn't seem like Audit is getting pulled into ActiveRecord properly

laserlemon:
Closing in favor of #102.

Issue #100, opened 2012-05-08
README Gemfile installation instructions don't work with rc gem


gbuesing:
The README shows the Gemfile setup as:

  gem "audited-activerecord", "~> 3.0"

Which doesn't find the gem on bundle install, because the latest version at Rubygems.org is 3.0.0.rc1.

Until 3.0 is released, the install instructions should be updated to:

  gem "audited-activerecord", "~> 3.0.0.rc1"

kennethkalmer:
Please accept my apologies for letting this one lie idle for so long. The README has since moved on with up to date instructions.

PR #102, opened 2012-05-11
Add alias to Audit class and fix README

carvil:
Hi,

This pull request aims at fixing the following issues:

- #101 - Add alias for the Audit class, otherwise one would have to use `Audited::Adapters::ActiveRecord::Audit`.

- #100 - I had exactly the same issue. I fixed the docs in order to reflect the latest released gem.

Do you agree with these changes?


tbugai:
This should be left up to the individual users to alias.

PR #112, opened 2012-08-13
Ensure controller isn't leaked in ActiveController around filter

braintreeps:
I found that in some cases the after callback was not being executed in audited. This had the side effect of leaking self.controller in tests and causing warden to show random failures. I changed the before/after callbacks to be an around callback with a begin/ensure block.

---
John Downey
Braintree Developer

### DIFF
--- a/lib/audited/sweeper.rb
+++ b/lib/audited/sweeper.rb
@@ -4,13 +4,13 @@ module Audited
 
     attr_accessor :controller
 
-    def before(controller)
-      self.controller = controller
-      true
-    end
-
-    def after(controller)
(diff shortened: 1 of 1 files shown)


kennethkalmer:
Merged this into a new branch (pr-112 158a6db33a6) for review by the others. Not sure exactly how to test it, but it is a legit fix, thanks so much @braintreeps.

/cc @danielmorrison @bkeepers

kennethkalmer:
Picking this up in #215 where we can have Travis look after us and get other kinds of feedback.

PR #119, opened 2012-10-05
Adds Rspec matchers to spec test your audited models

nhance:
The following matchers have been added:
- `be_audited`
- `have_associated_audits`

Usage:

Inside `spec_helper.rb` add the following:

``` ruby
require 'audited-rspec'
```

Then use in your models like this:

``` ruby
require 'spec_helper'

describe Company do
  it { should be_audited }
  it { should have_associated_audits }
end
```

``` ruby
describe User do
  it { should be_audited.associated_with(:company).except(:password) }
end
```

Closes reenhanced/audited#1

LGTM given by: @codenamev

### DIFF
--- a/lib/audited-rspec.rb
+++ b/lib/audited-rspec.rb
@@ -1,0 +1,4 @@
+require 'audited/rspec_matchers'
+module RSpec::Matchers
+  include Audited::RspecMatchers
+end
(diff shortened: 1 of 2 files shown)


drewbaumann:
This is super old, but took me too long to find. How would you feel about some of this information in the README?

PR #168, opened 2014-03-18
Fix: Destroy on unsaved object should not audit

rpond-pa:
If a record is new and destroy is called on it, an audit for the destroy should not be created.  It causes an ActiveRecord exception:

```
  ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved
```

This is a continuation of this pull request:  https://github.com/collectiveidea/audited/pull/79.  This version includes a (previously) failing spec.


kennethkalmer:
Thanks @rpond-pa for the contribution!

rpond-pa:
Woohoo!

PR #183, opened 2014-08-14
Support for around_audit callback

amysutedja:
In some scenarios, it might be useful for you to be able to have access to an immediately-created audit for a specific model. Ideally, you would pass your audit to the `after_audit` callback. However, the only way in which `ActiveSupport::Callbacks` supports access to the result of an executed block is via use of the `around` callback, which allows you to assign the result of the `yield` operation to retrieve the result of your action.

This changeset adds an `around_audit` callback to enable this scenario.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -65,10 +65,12 @@ module Audited
         before_update :audit_update if !options[:on] || (options[:on] && options[:on].include?(:update))
         before_destroy :audit_destroy if !options[:on] || (options[:on] && options[:on].include?(:destroy))
 
-        # Define and set an after_audit callback. This might be useful if you want
-        # to notify a party after the audit has been created.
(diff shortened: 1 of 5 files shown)


tquill:
Is there any documentation on how to use this callback?  I've added it to my model and it seems to break things, but I'm guessing I'm not using it properly.

amysutedja:
@tquill In my case, I needed to do something like the following.

```
def around_audit
  current_audit = yield

  Resque.enqueue(AuditWorker, current_audit.id)
end
```

Do you have some example code that you're expecting to work, but doesn't?

tquill:
That works!  I was making a dumb mistake by using the following from the code itself.  

```
def around_audit
  self.around_attr = yield
end
```

I'm not sure what I was thinking.  Anyway, thanks for the help.

PR #192, opened 2014-10-14
Handle serialized attributes to fix #191 

### LINKED ISSUE #191: Old revision of record with serialized attribute not correctly reconstructed
If an audited record has a `serialize` attribute that contains an array, then when creating an old revision of the record (using `audit.revision`), only one of the array elements is returned.

The problem is that `new_attributes` and `old_attributes` don't handle the case where an attribute may itself be an array.

k1w1:
Consider the action when reconstructing the attributes so that for 'create' and 'destroy' actions we don't mistakenly think the array in `audited_changes` contains an old/new value.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -78,7 +78,6 @@ module Audited
     def old_attributes
       (audited_changes || {}).inject({}.with_indifferent_access) do |attrs,(attr,values)|
         attrs[attr] = (action == 'update') ? values.first : values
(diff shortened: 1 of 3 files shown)


tbrisker:
This PR has been open for several years now with no activity and the main code has diverged from this branch. @k1w1, do you think it's still needed and would you be interested in returning to this or should we close it?

tbrisker:

PR #203, opened 2015-01-21
ensure enable back auditing on exception

mikz:
AuditedClassMethods#without_auditing would not enable auditing on exception

using ensure block will enable back even when exception is raised inside the block


kennethkalmer:
Thanks @mikz for the patch

PR #206, opened 2015-02-05
Rails 4.2 support

miks:
Can we make dedicated 4.2 release for rails 4.2?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Audited [![Build Status](https://secure.travis-ci.org/collectiveidea/audited.png
 
 **Audited** (previously acts_as_audited) is an ORM extension that logs all changes to your models. Audited also allows you to record who made those changes, save comments and associate models related to the changes.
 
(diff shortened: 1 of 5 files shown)


miks:
Fixed.

robmathews:
Tripping over this issue in rails 4.2 - w/o this fix, it doesn't work. Any plans to merge?

mordocai:
I'm currently working on moving to rails 4.2 and this is going to bite me. Would really appreciate it being merged ASAP.

robmathews:
My reaction to this was to switch to paper_trail.

It seems much more active and full featured. It has a "revert" function,
for example.

mordocai:
I feel a bit like a jerk discussing this in the issue on audited, but oh well. 

Unfortunately, I already have a considerable amount of production logs with audited so it'd be a real pain to migrate to something else.

I don't want to support audited myself though so if that's the other option I may take a look paper_trail.

I suppose I could always fork but I don't really want to maintain a persistent fork.

teknull:
I can see this being a big issue as we all migrate to 4.2.  I've had a lot of problems with the gem in Rails 3.2.x now that I'm upgrading multiple apps to 4.2 this concerns me.  Can we have compatibility and a gem release, please?

PR #208, opened 2015-02-13
Ready travis for the future

kennethkalmer:
I've done some work to get us on the path to merging the Rails 4.2.0 work (by hooking it up to CI), I also added tests for the upcoming JRuby-9000 release so we can be sure things work there. Also tried to make the tests faster by switching to Travis' container build system.

The big surprise for me though was that all our ActiveRecord tests we're running against in-memory sqlite3, and not against sqlite, mysql or postgres. Once I hooked this up I actually got a failing spec across the board from the thread-safety tests. The fix was easy, and luckily only the spec was affected.

The gemspec got a bit of TLC, took some hints from paper_trail *cough*.

So, any feedback before I pull the trigger?

/cc @danielmorrison


kennethkalmer:
@danielmorrison :+1: for merging, don't know how I missed the email notification...

PR #215, opened 2015-04-07
Ensure controller isn't leaked in ActiveController around filter (#112)

kennethkalmer:
> I found that in some cases the after callback was not being executed in audited. This had the side effect of leaking self.controller in tests and causing warden to show random failures. I changed the before/after callbacks to be an around callback with a begin/ensure block.

Originally from #112, but the source repo doesn't exist anymore.

I commented on the original PR that the fix is legit, but I need some advice on testing it (if testing is required).


cobbr2:
I'm merging this into our fork of 3.0-stable because it completely resolves an issue we have; perhaps how we repro'ed that would be helpful in testing. In our case, we audit our User model and are testing our SessionsController (logins). In a case where we tested for an "unactivated" user (i.e., an invalid one), ``Devise`` would ``throw :warden`` (and ``catch`` it). However, the next test case after that could not create a User model, because Audited would try to find the :current_user from the previous test case's controller's context - and in that context, the ``current_user`` would ``throw warden`` again.

PR #220, opened 2015-04-28
Rename revision accessor from :version to :audit_version

sellittf:
preventing clashing with existing attribute name


danielmorrison:
This is a great reference if someone needs to rename this. I don't think it helps others and may cause issues with existing apps, so I'm closing it.

PR #221, opened 2015-04-29
Add a Custom Auditor mention to the README

jdurand:
I scratched my head a bit on this one. I could only find my answer here https://github.com/collectiveidea/audited/blob/cd7065ef6fb9640a3a4e204477dd38dc0d12e50c/lib/audited/adapters/active_record/audit.rb#L40


kennethkalmer:
Thanks for the PR! May I suggest that you rename the helper method to `audited_user` and add another snippet of code showing how to configure Audited to call a different method?

```rb
Audited.current_user_method = :audited_user
```

PR #236, opened 2015-10-16
Evaluate column black-/whitelist lazily

NobodysNightmare:
## What changes?

With this PR the column selectors (`only` and `except`) are evaluated lazily on the first call of `non_audited_columns` instead of being evaluated immediately when calling `audited` (which happens by solely requireing the audited model).

## Why? What was the problem?

The reason for that change is that calling methods like `column_names` and `primary_key` require the database to be properly set-up. With an empty database those calls will fail and therefore it was impossible to require an audited model before its database table was setup properly, because the call to `audited` would immediately call those methods.

This does not seem to be a problem under normal, ActiveRecord-only conditions. However, we use both: Mongoid and ActiveRecord in our project.
During `rake db:setup` our Mongoid seems to require all files under `app/models` at some point, which will cause trouble, because when running `db:setup` there will (obviously) be no tables and therefore rake will fail as soon as one of the audited models is required.

By evaluating the excluded columns lazily, requiring a model will not cause any database interaction. I think this is a good thing in general, though I don't know if there are any ActiveRecord imposed rules on when you are allowed to call to the database.

## Note on target branch

Since we are using audited for a Rails 3 project I am issuing this PR against your 3.0 branch. Feel free to merge this PR whereever you see fit. I am not sure if you are even maintaining the 3.0 releases anymore. Still I would apprechiate if this could become part of Audited 3.x, since it would allow us to use a proper release again instead of relying on an own fork.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -47,17 +47,14 @@ module Audited
         # don't allow multiple calls
         return if self.included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
 
-        class_attribute :non_audited_columns,   :instance_writer => false
+        class_attribute :column_options,        :instance_writer => false
         class_attribute :auditing_enabled,      :instance_writer => false
         class_attribute :audit_associated_with, :instance_writer => false
 
-        if options[:only]
-          except = self.column_names - options[:only].flatten.map(&:to_s)
-        else
-          except = default_ignored_attributes + Audited.ignored_attributes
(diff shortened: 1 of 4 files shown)


NobodysNightmare:
Closing this PR, since #240 has been merged and achieves essentially the same thing. Some feedback would have been nice, though.

danielmorrison:
@NobodysNightmare sorry about the lack of feedback. I found 2 PRs that both needed some work after other merges had happened, and I tried to put them together as best as possible. 

Noted that feedback is good, and will work on that!

PR #240, opened 2015-11-07
Do not connect to db until needed fix #239

### LINKED ISSUE #239: Forced database connection on initialization
The call to primary_key on [auditor.rb#L88](https://github.com/collectiveidea/audited/blob/1f7724a1ab5a883f6dbe55e8c82abb7b0acffa0d/lib/audited/auditor.rb#L88) triggers a database connection when the class is initialized.

This breaks not only the Rails 4 convention, but also any environment that needs to load the application files without a database. Eg, if building a Docker image on a server where the database is not accessible, asset precompilation fails because the models that use Audited cannot be loaded.

agios:
We can put the calls that trigger a database connection in a proc and only call it when first needed.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -38,17 +38,19 @@ module Audited
         # don't allow multiple calls
         return if included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
 
-        class_attribute :non_audited_columns,   instance_writer: false
-        class_attribute :auditing_enabled,      instance_writer: false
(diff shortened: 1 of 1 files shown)


agios:
Just noticed #236 does essentially the same thing as this PR, albeit in a slightly different way.

But there are many cases where users might be affected by this, it would be nice if it is fixed, one way or another.

davidwessman:
:+1: Would really need this

dup2:
:+1 Please merge this, we can not precompile assets without it

dup2:
Great, thx.

PR #243, opened 2015-11-25
Remove audit#version

jkronz:
- remove calls to set the version number
- update scopes that rely on version number to find ancestor versions.

These changes only remove one feature that was supported by the parent gem: calling `obj.revision(:previous)` repeatedly. So recursing previous versions to find a change wont work anymore.  

```
while(some_revision_im_looking_for) do
  version = obj.revision(:previous)
end
``` 

will instead have to loop and check the revision index specifically:

```
revision_num = obj.audits.count
while(some_cond) do
  revision_num -= 1
  version = object.revision(revision_num)
end
```

I haven't tested this in our app yet, just removing the version column here and making the specs pass again.


jkronz:
sorry, made this request to the wrong fork. :cold_sweat:

PR #253, opened 2016-01-08
user_agent added to audits

ailinykh:


danielmorrison:
I like this in theory, but I don't want to put too much stuff into Audited that not everyone needs. Maybe `remote_address` and `request_uuid` might have already crossed that line.

An approach I'd rather see is an easy way to hook in to the sweeper to add additional data. That way everyone can add whatever they want.

PR #269, opened 2016-03-24
Make temporary disabling of auditing threadsafe

cluesque:
For compatibility with, e.g., sidekiq.

Otherwise, one thread disables audits other threads that should be auditing don't, until the first thread completes.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -39,7 +39,6 @@ module Audited
         return if self.included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
 
         class_attribute :non_audited_columns,   :instance_writer => false
-        class_attribute :auditing_enabled,      :instance_writer => false
         class_attribute :audit_associated_with, :instance_writer => false
 
         if options[:only]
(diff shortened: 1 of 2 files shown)


cluesque:
Note that this fails Travis because the `4.2-stable` branch fails, broken by collectiveidea/audited#235.

This branch does pass on my localhost.  Not sure how to convince travis to test it.

lucasmazza:
@danielmorrison @cluesque hey, there's any plans on going forward with this patch? We stumbled upon a similar scenario using Sidekiq and would be great to have a patch for making `without_auditing` thread safe.

lib/audited/auditor.rb
```diff
@@ -261,6 +271,18 @@ def enable_auditing
       def audit_as( user, &block )
         Audited.audit_class.as_user( user, &block )
       end
+
+      def auditing_enabled
+        if (val = Thread.current[const_get('AUDIT_VAR_NAME')]).nil?
```
lucasmazza: Going through the `audited` code I noticed that `Audited.store` exists for dealing with thread safe state, so maybe this check could be written as `Audited.store.fetch("#{name.tableize}_auditing_enabled", true)` or something similar to let the `Audited.store` method know about thread variables instead of having it here.

cluesque: Thanks @lucasmazza, that's much cleaner.

danielmorrison:
Looks good to me, but I don't like that the tests are blowing up. I won't have time to look at those today, so if anyone wants to check in the meantime…

danielmorrison:
Oh right, these were broken before because this is on the 4.2. stable branch. Let's merge and then figure it out.

PR #276, opened 2016-07-01
New release for Rails 5

danielmorrison:
Prepping for a new version (4.3.0) since Rails 5 is final. 

Mostly testing and README changes.

Still need: 

- [ ] [rails-observers](https://rubygems.org/gems/rails-observers) gem needs to release a new version to support Rails 5. In the meantime, you can point to the github repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ rvm:
   - 2.2.4
   - 2.3.1
   - ruby-head
-  - jruby-head
 env:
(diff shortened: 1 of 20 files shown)


freemanoid:
So the only working option for now is 
```
gem 'audited', github: "collectiveidea/audited"
gem "rails-observers", github: 'rails/rails-observers'
```
?

PR #279, opened 2016-08-01
Correct auditing_enabled for STI models

tbrisker:
The store used for setting and checking auditing_enabled uses
`name.tableize`, which causes STI models to check the incorrect key - if
auditing is disabled on the parent, children should also not be audited.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -285,11 +285,11 @@ module Audited
       end
 
       def auditing_enabled
-        Audited.store.fetch("#{name.tableize}_auditing_enabled", true)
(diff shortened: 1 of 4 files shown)


tbrisker:
@danielmorrison Thanks for the speedy merge! would it be possible to pull this back to the 4.2 branch and release a new version? This broke our CI and we are currently pinning to 4.2.0 to prevent it. If you wish I can submit another PR against 4.2-stable.

danielmorrison:
Working on it between meetings this morning. Should have it shortly.

danielmorrison:
4.2.2 is up.

PR #286, opened 2016-08-22
Normalise singular arguments for `on` option

benlovell:
It's idiomatic to normalise scalar arguments and I noticed this is the
case for the `except` option so I modified the handling of the `on`
option to suit. This means we can:

```ruby
class SomethingAudited < ActiveRecord::Base
  audit on: :create

  # would previously have required
  # audit on: [:create]
end
```

A small change but it adds some consistency and is less 'surprising'
perhaps.


danielmorrison:
Great idea!

benlovell:
😍

PR #288, opened 2016-08-23
Optional asynchronous audit creation

jimm:
Adds the option to create audit records asynchronously using a queueing framework such as Resque. See the new sections in the README: "Asynchronous Auditing" and those following it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,113 @@ To disable auditing on an entire model:
 User.auditing_enabled = false
 ```
 
+### Asynchronous Auditing
+
+To create audit records asynchronously, you need to tell Audited which
+adapter to use by setting `Audited.async_class`. Here is an example
+`config/initializers/audited.rb`:
+
+```ruby
+if Rails.env.test?
+  Audited.async_class = Audited::Async::Synchronous
+else
+  Audited.async_class = Audited::Async::Resque
+  # The default queue name is :audit. To specify your own, do this:
+  Audited::Async::Resque.queue = :my_queue
+end
+```
+
+In your model, set `async` to true:
+
+```ruby
+class User < ActiveRecord::Base
+  audited async: true
+end
+```
+
+There are two adapter currently available: `Audited::Async::Resque` and
+`Audited::Async::Synchronous`. The Resque adapter's default queue name is
+`:audit`. The latter is mostly used for testing, since it synchronously
+creates all audits passed to it.
+
+Using asynchronous auditing will trigger a deprecation warning in some
+versions of Rails related to the use of `after_commit`. The warning
+describes the issue and includes directions on how to opt in to the new
+behaviour and remove the warning:
+
+```ruby
+config.active_record.raise_in_transactional_callbacks = true
+```
+
+#### Behind the Scenes
+
+When an audit records should be created, its attributes are put into a
+class-level array. On commit, that array is sent to the async adapter for
+processing.
+
+If the adapter raises an error when trying to enqueue audits, the audits are
+written synchronously instead.
+
+#### Creating an Async Adapter
+
+Each Audited::Async adapter must implemented an `enqueue` class method that
+takes two arguments: the name of the audit class to instantiate and an array
+of audit creation attribute hashes. For each attribute hash, it should
+asynchronously create the audit record from the class and attributes. Here's
+an example:
+
+```ruby
+module Audited
+  module Async
+    class Resque
+      @queue = :audit
+
+      def self.enqueue(klass_name, audits_attrs)
+        Resque.enqueue(self, klass_name, audits_attrs)
+      end
+
+      # Takes a model `klass` and an array of hashes of audit `attrs` and
+      # creates audit records from them.
+      def self.perform(klass_name, audits_attrs)
+        klass = Module.const_get(klass_name)
+        audits_attrs.each do |attrs|
+          klass.create(attrs)
+        end
+      end
+    end
+  end
+end
+```
+
+When adding an adapter, make sure to add it to the list of autoloads in
+`lib/audited/audit.rb`.
+
+### Disabling Asynchronous Auditing
+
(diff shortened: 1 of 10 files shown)


danielmorrison:
Maybe add a mention of "why" in the README?

jimm:
As I wrote the most recent addition to the README, I realized that I'd thought about but not implemented something: passing the current time to the audit record so would could use that as the `created_at` timestamp. I'm going to add that right now.

danielmorrison:
Ha! Not going to lie, I saw that in the README and thought, "why doesn't he just add it?" 😄

jimm:
I obviously need more coffee. Sorry for all the little tweaky commits. I can squash them if that would be helpful.

danielmorrison:
Doesn't really bother me, so whatever you want to do. Now get that test suite passing! 😉 ☕

jimm:
`rake test` passes for me locally. What command should I be using to run the tests so I can see the failures?

jimm:
Ah, I see the problem in Travis:
```
4.0.13/lib/active_record/dynamic_matchers.rb:22:in `method_missing': undefined method `raise_in_transactional_callbacks=' for #<Class:0x00000004748de0> (NoMethodError)
```
That's from something I put in `spec/rails_app/config/environments/test.rb` to suppress a warning during testing. If I remove that then there might be warnings but it will probably fix this Travis failure. Shall I go ahead and do that and see what happens?

danielmorrison:
We're using [appraisal](https://github.com/thoughtbot/appraisal) so `appraisal install` and `appraisal rake` should get you there.

jimm:
Thanks. I could also add an `if respond_to?(raise_in_transactional_callbacks=)`. I'll see what happens with `appraisal`.

jimm:
Specs are passing.

ersatzryan:
I don't like this being part of audited core. I could see it as an `audited-async` gem that does what this PR does. If we were to merge this in we would then have to tackle the idea of being background processing gem agnostic as well as ensure audited's async features work with new releases of said background processing gems.

danielmorrison:
@ersatzryan I'm only comfortable with it in core if ansyc is *not* the default, but open to any ideas.

tbugai:
Is this solving a real problem?  Or are we solving it in the right way?  Depending on the background queue you use it is still going to create a record.

jimm:
As written, the existing behavior is the default; async is turned on by saying `audited async: true`. There is no dependency on any async library since `autoload` is used to bring them in if the user configures audited to use any particular background job gem.

@tbugai I understand that the problem this is solving (apps slowing down while waiting for audit records to be written) is not one that everybody faces. The whole point is to always create the audit record, just do it asynchronously. Or are you saying that some background queue gems will run synchronously?

albus522:
What is this actually solving? Queueing a async job vs saving a db record should be a very similar execution time.

You are effectively storing the audit information in a database so that you can again store it in a different database, while making it so that the canonical source of information is not accurate for some indeterminate amount of time that will vary based on load and what servers are involved.

jimm:
Testing with our application showed a 14% speedup with audit-intensive code. That is of course an anecdote, not data :-)

jimm:
Two more things:
- The audit information may not ever be stored to disk; it might be handed to a job that can execute it immediately.
- Even it the audit info is written to a database first by the background job, that can be done in a separate asynchronous thread which speeds up the main application.

I understand the issue @albus522 raises that the "canonical source of information [would] not [be] accurate for some indeterminate amount of time". That is not a requirement for audit records for the kinds of applications I write. I need to know what happened and when (and by whom), but don't need that information to be completely up to date as of the current moment. If being that current is a goal of the audited gem, then @albus522 is correct that this PR may not make sense.

However, since users of the gem would have a choice --- the default existing sync behavior or the new optional async behavior --- I don't see any harm in adding this to the gem. That is of course a decision only the gem devs can make.

getabetterpic:
+1 for this. Our use case is we are creating a model and adding a couple thousand associations to it, each of which is being audited. This causes the web request to time out since the audits are being handled synchronously. With async support, it would allow the audit transactions to take place in the background, thus freeing up the web request to return in a more reasonable amount of time.

ianks:
Would it not be preferable to just use ActiveJob for the first interface? This way, there is out of the box support for a number of adapters as opposed to just resque.

jimm:
That's a great idea, @iank. Unfortunately, we don't use ActiveJob and I don't have time to set up an environment that uses it and understand it well enough to make the change and test it properly.

In any case, I'm considering

jimm:
(Darn it. I hate hitting"Close" when I meant to simply cancel a comment I was writing. Sorry about that.)

@ianks That's a great idea. Unfortunately I don't use ActiveJob and don't have time to set up an environment to make that change and test it properly. Perhaps somebody else could fork my fork?

In any case, my fork is working well for us in our app. I'm hoping to have time to turn it into a separate gem so that we can continue to use the base Audited gem and its future upgrades instead of using this modification of the gem, but don't know if/when I'll be able to do that.

jimm:
Thank you everybody for discussing and considering this PR. It looks like it is not going to be accepted. Perhaps I'll turn it into a gem if I have spare time.

I'm going to close this PR. Thanks again for everyone's time and attention.

icecold21:
too bad, this PR is really good though.

jimm:
@icecold21 Thank you. I've recently created a much simplified solution that does not require modifying the gem itself. I may turn it into a separate gem, and just published a blog post about it this morning: http://dirtybirds.chloeandisabel.com/development/2017/05/03/async-audits-are-awesome.html

dan987:
@jimm link is dead would love to see solution.

jimm:
@dan987 sorry about the dead link. Here's a copy of that post I made to my personal blog: https://jimmenard.com/2017/05/03/async-audits-are-awesome.html

PR #292, opened 2016-09-13
Fix auditing instance attributes if "only" option specified

freemanoid:
Currently it's possible to add non-column attribute and it will be audited even though we have `only` options specified.
The problem is that we rely on `ActiveModel::Dirty#changed_attributes` which records non-column attributes but we excludes column attributes only.

This PR adds a test case and working solution.


freemanoid:
Very good, thank you! Any plans on making a release with this fix?

Issue #247, opened 2015-12-08
Avoid orphans on destroy to fix issues with MySQL


ledermann:
Currently I stumbled upon a serious issue using **audited** with MySQL. 

[MySQL doesn't manage persistent AUTOINCREMENT values for InnoDB tables](http://dev.mysql.com/doc/refman/5.7/en/innodb-auto-increment-handling.html#innodb-auto-increment-initialization): After a server restart, MySQL calculates the next value by doing a `SELECT MAX(id)+1 FROM table`. Because of this, a value can be reused if you delete the last inserted record. Since 10 years ago there is a [bugreport](https://bugs.mysql.com/bug.php?id=21641) for this.

This is a problem with **audited**, because in its current implementation it creates orphans if a record gets destroyed (audit record associated with a non-existing auditable / associated).

A possible solution could be:

1. Use `dependent: :nullify` for both `has_many` associations
2. Set `:auditable` to NULL for audits with `action == 'destroy'`

What do you think? Are you open for a PR to fix this issue?

ledermann:
Consider the following scenario to understand why the current implementation of the gem is fatal:

1. Assuming we have the model "Document" enabled for auditing. The database is _MySQL_, the ID field is defined as `int(11) NOT NULL AUTO_INCREMENT` (pretty standard)
2. Lots of documents are created, some of them were deleted. After adding Document with id 99, a document with id 100 was created and deleted afterwards. There are two Audit records for this document, both with `auditable_id == 100`. Both audits violate referential integrity, because there is no document with ID 100 anymore.
3. Now the MySQL server gets restartet (e.g. because of an software update). Because the AUTOINC status is not persistent, the MySQL server will recalculate the value for the documents table by doing a `SELECT MAX(id)+1 FROM documents`. It will be 99+1 == 100.
4. A new document is created. It will get the id 100, which means it **reuses** the id of a previously destroyed record. A new Audit record is created with `auditable_id == 100`

Now, `Document.find(100).audits.count == 3`, which is a **serious bug**, because the newly created document is linked with unrelated audits.

PR #294, opened 2016-09-20
Avoid orphaned audits

ledermann:
Leaving orphans is harmful. With MySQL it's a disaster, because record id's can be recycled (please see #247). But it's never a good idea to have foreign key references pointing to the nowhere.

This PR cleans up existing audits if an audited record gets destroyed by nullifying the references. This ensures referential integrity of the database.


danielmorrison:
One downside seems to be that you could no longer restore a record to its original state (with original id). That isn't a design goal of Audited, but I do think it is something people use it for.

ledermann:
@danielmorrison You are right, with my proposal the following scenario will **NOT** be possible anymore: 

1. Create a auditable (_audit 1_)
2. Change some, but not all attributes (_audit 2_)
3. Delete the auditable (_audit 3_)
4. Restore the auditable from _audit 2_. This does not work because _audit 1_ can not be found anymore

Fun fact: There is no spec for such a feature in the test suite :) 
My PR does not brake the tests. Ok, I have made some small adjustments, because `record.audits` now returns a blank list if the record is deleted.

IMHO, by all means issue #247 should be fixed (please see my [additional comment](/collectiveidea/audited/issues/247#issuecomment-248576386)). Otherwise, using the gem with MySQL can not be recommended.

If you don't like my fix, maybe we can look for another one. Without further investigations, an additional field in the audits table could help. On destroy, this field gets a uniq value (some kind of UUID or a timestamp) to link the audits together. But this would be a more complex change.

ledermann:
@danielmorrison Seems you are not convinced. What about about making the `after_destroy` callback optional via `options[:strict]` or something?

Please note, that restoring a record to its original state (with original id) will fail if the id is reused by another record.

danielmorrison:
Don't take my slow response as anything other than a slow response. ;) I probably won't have time to dig in until Monday or so.

PR #295, opened 2016-09-20
Set created_at as collection cache key

bmedenwald:
With Rails Conditional GET support (304 caching), the `stale?` and `fresh_when` methods rely on `collection_cache_key`, which by default utilize the `updated_at` column.  Because the audits table doesn't include an `updated_at` column, this throws an SQL error.

This change simply changed the default and relies on `super`.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -125,6 +125,11 @@ module Audited
       record
     end
 
+    # use created_at as timestamp cache key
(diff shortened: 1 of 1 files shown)


jaredbeck:
Would a better title for this PR be "SQL error in Rails Conditional GET (304 caching)"? Might help searchers in the future.

PR #297, opened 2016-09-22
Fix as_user documentation

andrewbredow:
`Audited.audit_class` in the correct way to access the audit class now.


danielmorrison:
That was the old way, but as of 4.3 you can do `Audited::Audit` directly.

PR #304, opened 2016-10-06
`json` and `jsonb` as possible `audited_changes` column type, fix #216

### LINKED ISSUE #216: Ability to change audited_changes column type to JSON format
Hi,

Would be nice to be able to switch `audited_changes` column to JSON format for PostgreSQL and be able to query inside this field like:

``` ruby
Audited::Adapters::ActiveRecord::Audit.where("(audited_changes->'name') IS NOT NULL")
```

Regards,

freemanoid:
- update migration
- set postgres travis version to 9.4 to allow `jsonb`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ env:
   - DB=SQLITE
   - DB=POSTGRES
   - DB=MYSQL
+addons:
+  postgresql: "9.4"
 gemfile:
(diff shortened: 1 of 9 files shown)


duboff:
Would be great to see this merged (only with JSONB instead of JSON)

domcleal:
Great, thank you @freemanoid!

PR #307, opened 2016-10-20
rename instance attribute version to audit_version

orenf:
seems dangerous to dynamically add an accessor method called version
because version is such a common name and could lead to some
unexpected behaviour in applications using audited.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -46,7 +46,7 @@ module Audited
     def revision
       clazz = auditable_type.constantize
       (clazz.find_by_id(auditable_id) || clazz.new).tap do |m|
-        self.class.assign_revision_attributes(m, self.class.reconstruct_attributes(ancestors).merge(version: version))
+        self.class.assign_revision_attributes(m, self.class.reconstruct_attributes(ancestors).merge(audit_version: version))
       end
     end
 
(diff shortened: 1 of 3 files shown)


domcleal:
I agree and the PR looks good, but I'm inclined to consider the `version` reader as part of the public API, so it should be deprecated first.

Would you be able to add this? (Perhaps using method_missing to avoid overwriting a user's `version` method?)

orenf:
Yeah, I think I can do that.

domcleal:
Thanks @orenf, please let me know when it's available to re-review.

orenf:
np @domcleal, I just need a few days to catch up on things.

diegosky1:
Anything I can help you guys with to make this happen?

tbrisker:
This PR has been open for over a year now with no activity. @orenf, do you think it's still needed and would you be interested in returning to this or should we close it? @diegosky1 would you be interested in picking this up if @orenf is no longer interested and addressing the previous comments?

jeffdill2:
@tbrisker @domcleal do you need someone to finish this one out, since it still seems like it'd be a good idea?

tbrisker:
@jeffdill2 Thanks, that would be awesome!

jeffdill2:
@tbrisker I finally got around to finishing this one out as well. 😃

New pull request here: https://github.com/collectiveidea/audited/pull/443

jeffdill2:
@tbrisker not a problem 👍

PR #317, opened 2016-12-07
Adding support for conditional audit writing.

azanar:
Adds support for an `if:` option to conditionally write audits.

The argument to `if:` is much the same as the argument to a Rails `if:` callback. It can be either:

- A symbol: meaning a method of the same name wants to be called, or
- A proc: which is called directly.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -219,9 +219,26 @@ module Audited
       end
 
       def write_audit(attrs)
+        return unless should_audit?(attrs)
         attrs[:associated] = send(audit_associated_with) unless audit_associated_with.nil?
         self.audit_comment = nil
         run_callbacks(:audit)  { audits.create(attrs) } if auditing_enabled
+      end
+
+      def should_audit?(attrs)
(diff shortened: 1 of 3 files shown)


lib/audited/auditor.rb
```diff
@@ -219,11 +219,28 @@ def audit_destroy
...
 
+      def should_audit?(attrs)
+        if audited_options[:if]
+          case audited_options[:if]
+          when Proc
+            audited_options[:if].call(self, attrs)
+          when Symbol
+            send(audited_options[:if], attrs)
+          else
+            errors.add(:audit, "Unsupported conditional argument of type #{audited_options[:if].class}")
```
danielmorrison: Seems like this should raise an exception instead of adding an error. Errors are user-facing and a user can't do anything about it. Exceptions are programmer-facing.

tbrisker:
This PR has been open for over a year now with no activity. @azanar , do you think it's still needed and would you be interested in returning to this or should we close it?

tbrisker:

PR #319, opened 2016-12-22
Allow custom Audit classes

### LINKED ISSUE #314: Why we don't have Audits model in rails ?
I am using audited gem and want some changes in Audit model like I want my Audit model to act as tenant is there any way I can access Audit model ?

nickrivadeneira:
Classes that inherit from `Audited::Audit` can be used instead of the
default by assigning `Audited.audit_class` in an initializer.

Resolves #314

### DIFF
--- a/spec/audited/audit_spec.rb
+++ b/spec/audited/audit_spec.rb
@@ -3,6 +3,54 @@ require "spec_helper"
 describe Audited::Audit do
   let(:user) { Models::ActiveRecord::User.new name: "Testing" }
 
+  describe "audit class" do
+    around(:example) do |example|
+      original_audit_class = Audited.audit_class
+
+      class CustomAudit < Audited::Audit
+        def custom_method
+          "I'm custom!"
+        end
+      end
+
+      class TempModel < ::ActiveRecord::Base
+        self.table_name = :companies
+      end
+
+      example.run
+
+      Audited.config { |config| config.audit_class = original_audit_class }
+      Audited::Audit.audited_class_names.delete("TempModel")
+      Object.send(:remove_const, :TempModel)
+      Object.send(:remove_const, :CustomAudit)
+    end
+
+    context "when a custom audit class is configured" do
+      it "should be used in place of #{described_class}" do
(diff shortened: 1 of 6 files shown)


nickrivadeneira:
One potential problem with this implementation is that it makes use of the `.audit_class` accessor that was previously deprecated. It fit so perfectly, so I used it, but I wonder if it'd break for people who were using it previously to assign adapters. Maybe not since those adapters have been removed and would result in a const missing exception that would need to be addressed anyway.

spec/audited/audit_spec.rb
```diff
@@ -3,6 +3,61 @@
...
+          audited
+        end
+
+        record = TempModel.create
+
+        audit = record.audits.first
+        expect(audit).to be_a CustomAudit
+        expect(audit.custom_method).to eq "I'm custom!"
+
+        Audited.config { |config| config.audit_class = nil }
```
domcleal: These cleanup sections should be in an `after` block, rather than the main spec body. (If possible, also move the new classes and setup to a `before` too.)

nickrivadeneira: @domcleal I've moved this into an `around` block so that the audit class can be reset to its original value. Let me know if you'd prefer the `before`/`after` split and having `config.audit_class` reset `nil` as it is here.

nickrivadeneira:
👍  Rebased. I made an additional change to `lib/audited/audit.rb` on lines 19 and 27 to ensure these changes take into account the addition of ce9bd0b975c09a4cf9e8a62da32d34ed7ff19ea8.

domcleal:
Thanks for the update @nickrivadeneira, merged!

VasudevaSharmaM:
Hi , 

I am using audited in building an application , I am facing the issue with this gem which says - "`method_missing': undefined method `audited_class_names' for #<Class:0x00007fffcef14138> (NoMethodError)", I am trying to work around and found that this behavior is expected from the previous conversation. Could anybody please assist me a solution for this ?

It would be very helpful

Thank you !!

PR #323, opened 2017-02-24
Allow update audits only for authenticated user

AlexKlim:
Sometimes we have a lot of background tasks, which update models, and we have a lot of useless records into DB. 

This PR allow track only if user is authenticated.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -250,7 +250,11 @@ module Audited
       end
 
       def skip_writing?
-        Audited.only_authenticated_user ? ::Audited.store[:current_controller]&.send(Audited.current_user_method).blank? : false
+        if Audited.only_authenticated_user
+          ::Audited.store[:current_controller] ? ::Audited.store[:current_controller].send(Audited.current_user_method).blank? : nil
+        else
+          false
+        end
       end
     end # InstanceMethods
 
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -219,6 +219,8 @@ module Audited
       end
 
       def write_audit(attrs)
+        return if skip_writing?
+
         attrs[:associated] = send(audit_associated_with) unless audit_associated_with.nil?
         self.audit_comment = nil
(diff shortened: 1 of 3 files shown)


lib/audited/auditor.rb
```diff
@@ -246,6 +248,10 @@ def auditing_enabled
       def auditing_enabled=(val)
         self.class.auditing_enabled = val
       end
+
+      def skip_writing?
+        Audited.only_authenticated_user ? ::Audited.store[:current_controller]&.send(Audited.current_user_method).blank? : false
```
domcleal: The `&.` operator is only available on Ruby 2.3 and higher, please check if the value's nil for compatibility with other versions.

spec/audited/auditor_spec.rb
```diff
@@ -604,4 +604,15 @@ def non_column_attr=(val)
       }.to_not change( Audited.audit_class, :count )
     end
   end
+
+  describe "Update only for authenticated user" do
+    before do
+      Audited.only_authenticated_user = true
```
domcleal: Also add an `after` block to reset this to the default value (`false`) to prevent it interfering with later tests.

AlexKlim: Done

lib/audited/auditor.rb
```diff
@@ -246,6 +248,14 @@ def auditing_enabled
       def auditing_enabled=(val)
         self.class.auditing_enabled = val
       end
+
+      def skip_writing?
+        if Audited.only_authenticated_user
+          ::Audited.store[:current_controller] ? ::Audited.store[:current_controller].send(Audited.current_user_method).blank? : nil
```
domcleal: 1. Could you use `store[:current_user]` instead? If the controller doesn't respond to the current user method then this would cause an error - the `:current_user` contains a lambda from Sweeper that properly checks this, you can re-use it for free. See `Audited::Audit#set_audit_user` for how to call it.
1. It should probably also `return false` if `store[:audited_user]` is set, because someone could call `Audited.as_user` even if the controller doesn't set the user.

PR #324, opened 2017-03-07
Docs: Add changelog entry for PR 295

jaredbeck:
[ci skip]

PR #325, opened 2017-03-16
Remove rails-observers dependency

domcleal:
The `Audited::Sweeper` class made limited use of the cache sweeping
functionality provided by rails-observer, mostly using it as a way to
implement `Audit#before_create` on the same class that was attached
around controller actions.

Since there is no released version of rails-observer that's compatible
with Rails 5.0 (and not in the foreseeable future without considerable
refactoring), the population of IPs, users and request UUIDs is replaced
by thread-local storage and the existing `Audit#before_create` methods.

This is a net code reduction and somewhat simplifies the "sweeper" by
removing the assumptions about how rails-observer works.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -16,7 +16,4 @@ end
 
 appraise 'rails50' do
   gem 'rails', '~> 5.0.0'
-
-  # The following needs to point to Github until the release of 0.1.3
-  gem 'rails-observers', github: 'rails/rails-observers', branch: 'master'
 end
--- a/README.md
+++ b/README.md
@@ -29,11 +29,6 @@ Add the gem to your Gemfile:
 gem "audited", "~> 4.3"
 ```
 
-If you are using rails 5.0, you would also need the following line in your Gemfile.
-```ruby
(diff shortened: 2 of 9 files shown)


thomasfedb:
Would be very keen to see this released.

thomasfedb:
This as well as #324 have been merged into my fork (https://github.com/thomasfedb/audited). If there's enough interest I'm happy to cut a new minor release, although I'll need to change the gem name.

domcleal:
Thanks @thomasfedb, I'd be interested in a new gem if audited itself or rails-observers aren't updated within a couple of weeks.

thomasfedb:
@domcleal no worries. If you think of a decent gem name (I'm thinking `trackback` or `metameta`), or if there's anything that needs fixing then drop me a PR or create an Issue. I also see it as an opportunity to refactor and drop support for Ruby < 2.4 and Rails < 5.

thomasfedb:
@domcleal also note that, from what I can see, rails-observers is fairly dead.

PR #327, opened 2017-03-25
Add `update_with_only_comment` option

jestemkarol:
Added on f2de4e606e97ec4c750dc9336d54cdffb88b7ab8
Currently there is a possibility to update an audit if only audit comment is present:

`assert_difference('post.audits.count', 1) { post.update_attributes(audit_comment: 'Only comment') }`

In my opinion, updating an audit with only comment given is purposeless, as in fact there are no changes at all

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ class User < ActiveRecord::Base
 end
 ```
 
+You can update an audit if only audit_comment is present. You can optionally add the `:update_with_comment_only` option set to `false` to your `audited` call to turn this behavior off for all audits.
+
+```ruby
+class User < ActiveRecord::Base
+  audited :update_with_comment_only => false
+end
+```
+
 ### Limiting stored audits
 
 You can limit the number of audits stored for your model. To configure limiting for all audited models, put the following in an initializer:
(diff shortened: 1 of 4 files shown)


domcleal:
Since this seems to be a matter of opinion, would you mind implementing this as a configuration option on `Audited` to control the behaviour?

jestemkarol:
I've updated PR with an implementation of the configuration option, but it seems like Travis cached bundler and it needs to be rebuilt somehow, I guess.

PS. Sorry for late response 😄

domcleal:
PR looks good to me! :+1:

I've opened a PR to fix the test failures, it's been a while since they ran. Once we've got that fixed then we can merge back and get this PR in.

> PS. Sorry for late response 😄

No worries, I'm sure I've done worse 😁

domcleal:
That's now fixed - would you mind doing `git merge origin/master` (or `git rebase origin/master` if you prefer) and pushing please? That'll pick up the test fixes I hope.

jestemkarol:
Yup, all green right now 👌

domcleal:
Thanks @jestemkarol 👍

Marri:
@jestemkarol thank you for this! We had hacked around this and our code is much cleaner now.

PR #330, opened 2017-03-27
Fix deprecated spec format

GeekOnCoffee:
DEPRECATION WARNING: Using positional arguments in functional tests has been deprecated,
in favor of keyword arguments


domcleal:
Thanks for the PR @GeekOnCoffee. Unfortunately the new syntax is incompatible with Rails 4, so I've put in a version conditional as part of https://github.com/collectiveidea/audited/pull/329/commits/8b81f3a383aeeb851b018bdf446d6af5f568253e#diff-a9f03059ecd9d98a8c52687bfc100aaaR79.

PR #332, opened 2017-03-29
Fix AR/Rails gemspec dependencies to permit 5.1

domcleal:
Sorry @danielmorrison, a daft error I only noticed when viewing the published gem at rubygems.org.


danielmorrison:
Good call.

danielmorrison:
Released.

domcleal:
Thanks @danielmorrison!

PR #333, opened 2017-04-03
Support UUID primary keys for User

zshannon:
We use `uuid` as our primary key type on `users`, so we needed to tweak the install migration and are submitting back a generator option here.

If this is acceptable, I'm happy to add usage instructions to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ $ rails generate audited:install
 $ rake db:migrate
 ```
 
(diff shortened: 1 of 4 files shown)


domcleal:
Yep, that's fine - please do document it and add a test to `test/install_generator_test.rb`.

zshannon:
Thanks! Let me know if there's more I can do.

domcleal:
Merged, thanks @zshannon!

PR #335, opened 2017-04-04
audited_changes - used .type instead of .sql_type to serialize - support for oracle

NileshPanchal:


domcleal:
Merged, thanks @NileshPanchal!

PR #339, opened 2017-04-09
Allow redacting specific attributes from audits

tbrisker:
This allows users to define attributed which should be audited when
changed without saving the actual change in the audit log - for example
for passwords or secret keys.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -31,6 +31,10 @@ module Audited
       #       audited except: :password
       #     end
       #
+      # * +redacted+ - Changes to these fields will be logged, but the values
+      #   will not. This is useful, for example, if you wish to audit when a
+      #   password is changed, without saving the actual password in the log.
+      #
       # * +require_comment+ - Ensures that audit_comment is supplied before
       #   any create, update or destroy operation.
       #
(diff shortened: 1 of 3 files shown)


lib/audited/auditor.rb
```diff
@@ -138,10 +155,6 @@ def non_audited_columns
 
       protected
 
-      def non_audited_columns
```
tbrisker: Not directly related to this change but this function is just a duplicate deceleration of the one above it.

lib/audited/auditor.rb
```diff
@@ -81,9 +85,19 @@ def has_associated_audits
       def default_ignored_attributes
         [primary_key, inheritance_column]
       end
+
+      def normalize_audited_options(options)
```
tbrisker: normalizing the options on initialization simplifies the use later.

lib/audited/auditor.rb
```diff
@@ -81,9 +85,19 @@ def has_associated_audits
...
+        [:only, :except, :redacted].each do |opt|
+          options[opt] = Array.wrap(options[opt]).map(&:to_s)
+        end
+        self.audited_options = options
+      end
     end
 
     module AuditedInstanceMethods
+
+      REDACTED = '[REDACTED]'
```
domcleal: The hard-coded string replacement (and name of the feature) concerns me a little, because it's invalid for different field types (e.g. integer or JSON fields) and might not be correct for all users.

Would you be able to implement a more generic "munge" interface, that might let a user handle different types or fields or enter their own text per-field? Perhaps something as simple as a lambda that's passed a hash of attributes.

ares: @tbrisker what's the status, the PR would be super useful I think

tbrisker:
Closing this PR for now.

PR #342, opened 2017-04-19
Allow multiple relations;

gabrielms:
Allow multiple relations sending them as array of hashes.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -219,9 +219,12 @@ module Audited
       end
 
       def write_audit(attrs)
-        attrs[:associated] = send(audit_associated_with) unless audit_associated_with.nil?
-        self.audit_comment = nil
-        run_callbacks(:audit)  { audits.create(attrs) } if auditing_enabled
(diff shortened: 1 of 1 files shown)


lib/audited/auditor.rb
```diff
@@ -219,9 +219,12 @@ def audit_destroy
       end
 
       def write_audit(attrs)
-        attrs[:associated] = send(audit_associated_with) unless audit_associated_with.nil?
-        self.audit_comment = nil
-        run_callbacks(:audit)  { audits.create(attrs) } if auditing_enabled
+        audits_associated = audit_associated_with.is_a?(Array) ? audit_associated_with : [audit_associated_with]
+        audits_associated.each do |audit_associated|
```
domcleal: Won't this create multiple, duplicate audits, one for each associated object? Though it's more complex, I think it'd make more sense for one audit to exist per event, but to change to the schema to have multiple associations.

gabrielms: Hello @domcleal.

You are probably right. I had no time to take a better look and study the audited gem. I've made this because i needed in a project of mine.
It worked but i agree with you that this is not the right way of doing it. :(

PR #344, opened 2017-04-21
Call controller's current_user method during audited model creation

### LINKED ISSUE #336: audit.user no longer stored on 4.4.1
After updating audited from 4.3.0 to 4.4.1 `user` is no longer stored in audited record which is supposed to be `current_user` from controller

domcleal:
Restores behaviour prior to 87d402a where the sweeper would call the
controller's current_user method from the audited model callback.

Since 87d402a, the current_user method was called from an around action
callback registered on the base controller which was being called prior
to other callbacks that were authenticating the user. This caused
problems in apps where the user hadn't yet been set (so audit users were
nil), or CSRF issues because current_user was called prior to session
changes.

Fixes #336

### DIFF
--- a/lib/audited/sweeper.rb
+++ b/lib/audited/sweeper.rb
@@ -18,7 +18,7 @@ module Audited
     end
 
     def current_user
-      controller.send(Audited.current_user_method) if controller.respond_to?(Audited.current_user_method, true)
+      lambda { controller.send(Audited.current_user_method) if controller.respond_to?(Audited.current_user_method, true) }
     end
 
     def remote_ip
--- a/spec/audited/sweeper_spec.rb
+++ b/spec/audited/sweeper_spec.rb
@@ -1,6 +1,8 @@
 require "spec_helper"
 
 class AuditsController < ActionController::Base
+  before_action :populate_user
+
   attr_reader :company
 
   def create
@@ -17,6 +19,8 @@ class AuditsController < ActionController::Base
 
   attr_accessor :current_user
   attr_accessor :custom_user
+
+  def populate_user; end
 end
 
 describe AuditsController do
@@ -67,6 +71,18 @@ describe AuditsController do
       post :create
 
       expect(controller.company.audits.last.request_uuid).to eq("abc123")
+    end
+
+    it "should call current_user after controller callbacks" do
+      expect(controller).to receive(:populate_user) do
+        controller.send(:current_user=, user)
+      end
+
+      expect {
+        post :create
+      }.to change( Audited::Audit, :count )
+
+      expect(controller.company.audits.last.user).to eq(user)
     end
 
   end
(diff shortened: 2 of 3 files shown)


spec/audited/sweeper_spec.rb
```diff
@@ -1,6 +1,12 @@
 require "spec_helper"
 
 class AuditsController < ActionController::Base
+  if Rails::VERSION::MAJOR == 4
+    before_filter :populate_user
```
grosser: this will cause deprecations on 4.2 ... and I raise on deprecations ... so kaboom :D 
... also add the raise on deprecations option to find this in tests easier.

lib/audited/sweeper.rb
```diff
@@ -18,7 +18,7 @@ def around(controller)
     end
 
     def current_user
-      controller.send(Audited.current_user_method) if controller.respond_to?(Audited.current_user_method, true)
+      lambda { controller.send(Audited.current_user_method) if controller.respond_to?(Audited.current_user_method, true) }
```
grosser: just store the controller here and don't create a new lambda ?
... then move the remaining code to the sweeper

spec/audited/sweeper_spec.rb
domcleal: The deprecation warning was added in 5.0 - in 4.2 the use of before_filter was just ["discouraged"](http://guides.rubyonrails.org/4_2_release_notes.html#action-pack-notable-changes) so this should be fine whether AS deprecations are raised or not.

lib/audited/sweeper.rb
domcleal: Yeah, I was initially planning to store the controller but I didn't think it was a good idea to have the Audit model calling methods directly on the controller. I preferred to keep the controller-specific calls and logic in this class.

> ... then move the remaining code to the sweeper

Which code's that? Do you mean to the model?

grosser:
nvm, already answered it :)
either way is fine, creating an extra lambda seemed a bit weird to me, but
if you want to keep controller code the sweeper idk how else to do it

domcleal:
Thanks for the review @grosser, appreciated.

spec/audited/sweeper_spec.rb
lucianosousa: `before_action` already exists on rails 4.0
https://github.com/rails/rails/blob/f53132deea65ae673b29b811c606be103f1ae3ab/actionpack/lib/abstract_controller/callbacks.rb
I would suggest to remove this `if` here

domcleal: Ah, you're right - thanks for the info. I've updated the PR to remove the conditional entirely.

morgoth:
Can we have a release with this fix?

domcleal:
@danielmorrison I opened #345 with updates for 4.5.0 if you'd kindly push the gem.

morgoth:
@domcleal @danielmorrison hey, any news on releasing the new version?

domcleal:
audited 4.5.0's released.

morgoth:
Great, thanks :)

PR #351, opened 2017-05-22
Allowing using symbol keys in `non_audited_columns`

metalelf0:
Right now, even though the README file mentions symbol keys in `non_audited_columns`, they don't get filtered out as expected. To get `non_audited_columns` working as expected you have to pass the attribute names as strings.

This PR allows passing symbol keys.


domcleal:
Thanks @metalelf0!

PR #353, opened 2017-06-03
Fix change of  vritual attribute can not be tracked.

bealking:
Some gems like "acts-as-taggable-on" may add some vritual attribute/method that inherits activerecord dirty feature. Please see the follow links:

[https://github.com/mbleigh/acts-as-taggable-on/blob/64ed839a8a6021c719fbff9c6bdf74a4fcd65f18/lib/acts_as_taggable_on/taggable/core.rb#L193](https://github.com/mbleigh/acts-as-taggable-on/blob/64ed839a8a6021c719fbff9c6bdf74a4fcd65f18/lib/acts_as_taggable_on/taggable/core.rb#L193)
[https://github.com/mbleigh/acts-as-taggable-on/blob/8e64c3d4a81cfbb8af621228eae36a65c1f94501/lib/acts_as_taggable_on/taggable/dirty.rb#L14](https://github.com/mbleigh/acts-as-taggable-on/blob/8e64c3d4a81cfbb8af621228eae36a65c1f94501/lib/acts_as_taggable_on/taggable/dirty.rb#L14)

[https://github.com/collectiveidea/audited/blob/master/lib/audited/auditor.rb#L187](Audited::Auditor#audited_changes) uses rails "[](attr_name)" method to fetch the new value of changed attribute which leads those gems that didn't extend this method to return nil (please see the test example added).

so I change the method to public_send to solve this problem.

### DIFF
--- a/spec/audited/auditor_spec.rb
+++ b/spec/audited/auditor_spec.rb
@@ -180,6 +180,12 @@ describe Audited::Auditor do
       expect(@user.audits.last.audited_changes).to eq({ 'name' => ['Brandon', 'Changed'] })
     end
 
+    it "should store the change of the custom attribute which extended dirty feature" do
+			@user = Models::ActiveRecord::UserDelegateCompany.create(name: 'Brandon', username: 'brandon', password: 'password')
+      @user.update_attributes company_list: 'apple,google'
+      expect(@user.audits.last.audited_changes).to eq({ 'company_list' => ['', 'apple,google'] })
+    end
+
     it "should store audit comment" do
       expect(@user.audits.last.comment).to eq('Update')
     end
--- a/spec/support/active_record/models.rb
+++ b/spec/support/active_record/models.rb
@@ -13,6 +13,52 @@ module Models
       end
     end
 
+    class DelegatedCompany < ::ActiveRecord::Base
+      self.table_name = :companies
+      belongs_to :user, class_name: "UserDelegateCompany"
+    end
+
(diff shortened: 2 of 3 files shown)


spec/audited/auditor_spec.rb
```diff
@@ -180,6 +180,12 @@ def non_column_attr=(val)
       expect(@user.audits.last.audited_changes).to eq({ 'name' => ['Brandon', 'Changed'] })
     end
 
+    it "should store the change of the custom attribute which extended dirty feature" do
+			@user = Models::ActiveRecord::UserDelegateCompany.create(name: 'Brandon', username: 'brandon', password: 'password')
```
domcleal: Use two spaces, not tabs please.

spec/support/active_record/models.rb
```diff
@@ -13,6 +13,52 @@ def name=(val)
...
+      end
+
+      def save_company_list
+        @company_string.split(',').each do |name|
+          Company.find_or_create_by(name: name, owner_id: self.id)
+        end
+      end
+
+			# mock activerecord dirty feature
+			def company_list_changed?
```
domcleal: tabs/spaces

spec/support/active_record/models.rb
```diff
@@ -64,6 +110,7 @@ class Company < ::ActiveRecord::Base
     class Company::STICompany < Company
     end
 
+
```
domcleal: Unnecessary?

PR #355, opened 2017-06-16
display class and id for deleted users

grosser:
was about to make a PR when I realized it's not that easy ... since user should behave like a AR and most likely not return a weird string ... maybe user_as_string should do that instead and user should be aliased to something else ... idk ... 
/fyi @danielmorrison

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -117,7 +117,7 @@ module Audited
 
     # @private
     def user_as_string
-      user_as_model || username
+      user_as_model || username || ("#{user_type}##{user_id}" if user_type && user_id)
     end
     alias_method :user_as_model, :user
     alias_method :user, :user_as_string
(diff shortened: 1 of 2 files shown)


danielmorrison:
I'm ok rethinking that behavior (as long as we don't have breakages). What's the goal you're going for here?

grosser:
I wanted to show something in the UI when a user got deleted ... atm I'm
solving it in the view, but I though something like user_as_string should
try as hard as possible to display something ... on the other hand it could
be a security issue to leak user ids ... so more of a 'what do you think'
PR ... solved for me now so I can go either way

tbrisker:
Is this still wip @grosser?

grosser:
updated and ready to merge 🎉

grosser:
hmmm tests failed, might need more rework ... I guess I'll just close this until it bothers me again ...

Issue #359, opened 2017-07-07
Typo in RSpec Matchers


hernanat:
in `audited/lib/audited/rspec_matchers.rb`, I believe line 120:

    expects "non audited columns (#{model_class.non_audited_columns.inspect}) to match (#{expect})"

should be

    expects "non audited columns (#{model_class.non_audited_columns.inspect}) to match (#{except})"

Error I've been receiving:

    undefined local variable or method 'expect' for <Audited::RspecMatchers::AuditMatcher:0x007f8627d6ae48>

hernanat:
opened PR here: https://github.com/collectiveidea/audited/pull/360

hernanat:
resolved in #360

PR #360, opened 2017-07-07
Fix typo in Audited RSpec Matchers

hernanat:
Currently, the audited RSpec matchers throw an error:
`undefined local variable or method 'expect' for <Audited::RspecMatchers::AuditMatcher:0x007f8627d6ae48>`
when you use `.only` or `.except`.

The issue was a typo where it was referring to a non-existent variable `expect` instead of `except`.

Response to: https://github.com/collectiveidea/audited/issues/359


m-13:
@danielmorrison Is this good to merge, or does it need any changes?

PR #364, opened 2017-07-19
Test Ruby 2.2.2

sjreich:
I'm hoping that this is enough to run Travis against Ruby 2.2.2.  

Rspec tests pass locally on 2.2.2.

Please advise if more is required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 cache: bundler
 rvm:
   - 2.1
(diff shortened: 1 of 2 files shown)


tbrisker:
Hi @sjreich, any reason that we should test on that specific version as well? AFAIK there shouldn't be any significant difference between that and any other 2.2.x version.

sjreich:
We no longer need this.

PR #365, opened 2017-07-26
Remove duplicated method

printercu:
It's redefined on L141.


printercu:
I can do this. But this one is protected now, so one can use it only with `send`.

tbrisker:
Thanks @printercu !

PR #368, opened 2017-08-09
Proper single-table inheritance support for Audit#revision

noiges:
## Details
This PR fixes an issue where the subclass of a record from a single-inheritance table can't be recovered anymore with `Audit#revision` after the record was deleted.

The reason for this is that the `inheritance_column` is not audited by default, which means the subclass is impossible to retrieve after a record has been deleted.

My naive solution for this problem is to just track changes to the `inheritance_column`, and make sure the audited class is instantiated properly when a revision is created.

## Concerns
1. I am not fully aware why the `inheritance_column` has been ignored so far — this fix is based on the  assumption that this was just an oversight. Are there any issues with this? All tests are passing on my machine though.

2. If the inheritance column or the class name have ever been refactored without a migration of the existing audits’ inheritance column, an `ActiveRecord::SubclassNotFound` will be raised if an audit's revision is retrieved ([because we rely on this](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/inheritance.rb#L218)). Should this be handled more gracefully, with a fall back to the parent class?

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -64,9 +64,18 @@ module Audited
     # the object has been destroyed, this will be a new record.
     def revision
       clazz = auditable_type.constantize
-      (clazz.find_by_id(auditable_id) || clazz.new).tap do |m|
-        self.class.assign_revision_attributes(m, self.class.reconstruct_attributes(ancestors).merge(version: version))
-      end
+      attributes = self.class.reconstruct_attributes(ancestors).merge(version: version)
+      inheritance_column = clazz.inheritance_column.to_s
+
+      instance = if clazz.exists?(id: auditable_id)
+                   clazz.find_by_id(auditable_id)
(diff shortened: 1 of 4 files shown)


noiges:
Looks like all Rails 4 builds are failing. I'm gonna close this for now until I've figured out a fix.

noiges:
Turns out the specs were failing for Rails 4 applications that are using `protected_attributes`, since the inheritance column is blacklisted by default for mass assignment, which results in the parent class being instantiated, instead of the right subclass. I think this is fine, since it's consistent with ActiveRecord's default behaviour.

I've modified my specs to take this into account, for cases where `protected_attributes` are used.

januszm:
Can we add a check for this new code, like `if rails version >= 7.1` or similar and proceed with this code? I'd expect not have to add the `audited` code to all subclasses of a class that already defined `audited`.

PR #378, opened 2017-09-19
获取用户Ip

liu7899:


tbrisker:
Hi @liu7899 , why is this needed? Rails calculates the `remote_ip` in a manner that takes into account various possible connection methods, such as proxies - see https://github.com/rails/rails/blob/5-1-stable/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L92 for more info.

tbrisker:
Thank for your effort @liu7899, I'm going to go ahead and close this PR since it seems that remote IP calculation should be handled correctly by Rails already. 
If you think there is a problem with the way Rails calculates remote IPs, please open a PR against Rails itself. 
If there is an issue with the way we calculate the remote IP here, feel free to reopen this PR and explain why this change is needed.

PR #381, opened 2017-09-21
Add support for undoing changes

arcticbarra:
I decided to create this PR as in our implementation of audited we needed to undo certain changes.
- Destroy models that were created
- Create models that were destroyed
- Change back attributes that were updated


acrogenesis:
👍

PR #377, opened 2017-09-14
Use updated AR::Dirty API to find changed attributes

ulfmagnetics:
Fixes https://github.com/collectiveidea/audited/issues/371

PR #382, opened 2017-09-26
Fix deprecation warnings caused by using changed_attributes

### LINKED ISSUE #371: Deprecation warning for changed_attributes with rails 5.1
Calling `update_attributes` in my model when using audited shows the following deprecation warning

> #### DEPRECATION WARNING: The behavior of `changed_attributes` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_changes.transform_values(&:first)` instead.

I believe this is coming from https://github.com/collectiveidea/audited/blob/f8433d47cf998cd7acb12b2337bd5cddca643947/lib/audited/auditor.rb#L181

steves:
This fixes #371 

I've been running this locally on my own installation and everything seems to be working fine. The new `attributes_in_database` method is meant to be a drop in replacement for `changed_attributes` so there shouldn't be any problems with it.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -178,9 +178,9 @@ module Audited
         collection =
           if audited_options[:only]
             audited_columns = self.class.audited_columns.map(&:name)
-            changed_attributes.slice(*audited_columns)
+            attributes_in_database.slice(*audited_columns)
           else
-            changed_attributes.except(*non_audited_columns)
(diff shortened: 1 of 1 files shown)


JasonBarnabe:
This is the same change as in #377, but that one looks like it's backwards compatible.

Issue #321, opened 2017-02-06
Return nil on invalid revision number


clothesline0909:
When I try to access a revision that doesn't exist e.g. model.revision(n), where n is larger than model.revisions.length, I expect the method to return nil or an exception; however it just returns the last revision.

danielmorrison:
Something we never thought of. I'd be open to a PR with this.

JonathanWThom:
Hello, it doesn't look like a PR has been made for this yet. I'm going to see if I can come up with a fix, if no one else has yet.

tbrisker:
This has been resolved in https://github.com/collectiveidea/audited/commit/7c4fb94fe6df1a8ab655a689e2062ee496c9241e

PR #384, opened 2017-10-21
Return nil on invalid revision number

JonathanWThom:
* Previously, when a `record.revision(version)` was called
  with a version number that was greater than the total
  number of revisions, it would return the last revision
  in the list. This changes it to return nil.

This addresses this issue raised here > https://github.com/collectiveidea/audited/issues/321.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -119,7 +119,7 @@ module Audited
       # Get a specific revision specified by the version number, or +:previous+
       # Returns nil for versions greater than revisions count
       def revision(version)
-        if version == :previous || revisions.count >= version
+        if version == :previous || self.audits.last.version >= version
           revision_with Audited.audit_class.reconstruct_attributes(audits_to(version))
         end
       end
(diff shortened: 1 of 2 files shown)


JonathanWThom:
Perhaps there is a more elegant solution, or something I'm missing. But this seems to work and didn't blow up any tests (locally).

lib/audited/auditor.rb
```diff
@@ -117,8 +117,11 @@ def revisions(from_version = 1)
       end
 
       # Get a specific revision specified by the version number, or +:previous+
+      # Returns nil for versions greater than revisions count
       def revision(version)
-        revision_with Audited.audit_class.reconstruct_attributes(audits_to(version))
+        if version == :previous || revisions.count >= version
```
tbrisker: This will cause all revisions to be loaded from the database just so they can be counted, which can be very slow when an object had many changes.
How about comparing the version to `self.audits.last.version`?

JonathanWThom: @tbrisker That's a great idea, I will make that change.

tbrisker: Thanks @JonathanWThom ! While you're at it, can you please also add a line to the Changelog with the change?

PR #385, opened 2017-11-09
Columns order adjustment for polymorphic indexes

olegykz:
Current auditable & associated polymorphic indexes are using suboptimal columns order. PR for fixing order according to current Rails behavior in context of index structure for polymorphic association (https://github.com/rails/rails/commit/9cdd0a1fdf8308985231242d378e3a1c29c4ab00)


josephpage:
Why the user index was not changed too ?

tbrisker:
@josephpage PRs welcome :wink:

PR #386, opened 2017-11-13
update travis config to latest rubies

lostapathy:
Updating travis config to cover latest ruby versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ matrix:
       gemfile: gemfiles/rails40.gemfile
     - rvm: 2.4.2
       gemfile: gemfiles/rails41.gemfile
+    - rvm: ruby-head
+      gemfile: gemfiles/rails40.gemfile
+    - rvm: ruby-head
+      gemfile: gemfiles/rails41.gemfile
   fast_finish: true
 branches:
   only:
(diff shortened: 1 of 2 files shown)


lostapathy:
Also skip testing rails-4.x with ruby-head.  We're not testing them with ruby-2.4 and they fail anyway.

.travis.yml
```diff
@@ -2,9 +2,9 @@ language: ruby
 cache: bundler
 rvm:
   - 2.1
-  - 2.2.4
-  - 2.3.1
-  - 2.4.1
+  - 2.2.8
+  - 2.3.5
+  - 2.4.2
```
tbrisker: maybe just drop the z part so we're always testing with the latest version of each line?

lostapathy: I'm ok with that but went this way to leave the project's existing style alone.  Some projects actively don't like to do it that way.  Occasionally it seems that Travis gets the logic behind that wrong - i.e., for a while this summer 2.2 was getting you 2.2.1 not 2.2.8 like you'd expect.

tbrisker: @danielmorrison any preference?

tbrisker:
@lostapathy Thanks, I pushed a rebase to your branch to fix the merge conflicts, will merge once travis is 🍏

PR #387, opened 2017-11-29
Use Rails 5.1.4 instead of Rails 5.1.0.rc1

mrhead:
The format of gemfiles was changed by `bundle exec appraisal install`.


tbrisker:
This also fixes the failing CI on other PRs (5.2.0beta2 is picked up previously for rails51.gemspec). @danielmorrison can you merge please?

PR #388, opened 2017-11-29
Add tests for Rails 5.2

mrhead:
Rails 5.2 is imminent so we should add support for it.


tbrisker:
@mrhead looks like now tests for 5.1 fail on mysql, care to check why?

tbrisker:
disregard my comment, looks like #387 fixes this issue - apparently 5.2.0.beta2 < 5.2 causing the failures in CI.

mrhead:
@tbrisker Awesome :)

mrhead:
Anyway, just let me know if there is any action needed. I'll be happy to check it.

Issue #394, opened 2017-12-14
`revision_at` and related methods should work backwards instead of forward


RavenXce:
The current implementation results in all models which have `audited` added after it's creation to not give accurate revisions. 

This is as the model is reconstructed by finding the current model, and applying revision 1 upward to the required revision. However, it is not guaranteed that revision 1 is a `create` event and has the original set of attributes. This causes a situation where calling `revision(n)` is not a pure function which is very strange for an audit library.

A more sane strategy would be use the current or greatest revision and work backwards to find the state at the required revision. This would ensure that the best-effort state of the model at the required revision can always be generated correctly.

jason-o-matic:
It seems like this gets more complicated when you have Audited configured to store a limited number of audits, but I agree the current behavior is not ideal.

PR #395, opened 2017-12-14
Traverse audits backwards instead of forwards in time

RavenXce:
Fixes https://github.com/collectiveidea/audited/issues/394


RavenXce:
This breaks backwards compatibility with `#revision_at` for one test case - we give the best-effort version of the model (as explained in the issue) instead of `nil`.

lostapathy:
I'm running into an issue that would be resolved by this change of approach.  If this is updated to resolve the merge conflicts and travis issues, is there interest in merging it?

If we don't like this approach, I think we should explore a method of creating a "create audit" when auditing is added to existing models so that revision_at is usable.

PR #397, opened 2018-01-07
Normalize options on initialization and minor cleanup

tbrisker:
Normalizing options passed on audited initialization simplifies code in
further steps. Also cleaned up the code a bit.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -38,26 +38,29 @@ module Audited
         # don't allow multiple calls
         return if included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
 
+        extend Audited::Auditor::AuditedClassMethods
+        include Audited::Auditor::AuditedInstanceMethods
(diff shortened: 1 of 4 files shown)


lib/audited/auditor.rb
```diff
@@ -302,6 +284,18 @@ def auditing_enabled
...
+
+      protected
+      def default_ignored_attributes
+        [primary_key, inheritance_column] + Audited.ignored_attributes
+      end
+
+      def normalize_audited_options
+        audited_options[:on] = Array.wrap(audited_options[:on])
+        audited_options[:on] = [:create, :update, :destroy] if audited_options[:on].empty?
+        audited_options[:only] = Array.wrap(audited_options[:only]).map(&:to_s) if audited_options[:only]
```
tbrisker: We want to keep `audited_options[:only]==nil` if it wasn't specified since we use it to determine which columns to save.

Issue #379, opened 2017-09-19
`revisions` does too much unnecessary sql queries


Akhilesh05:
The `revisions` method does too much unnecessary sql queries, for each audit version it does 2 queries.
This method can actually be rewritten causing only 2 sql queries.
I have forked and modified the gem for me but I want you guys to look into the matter. I have a large system and I need audit but I cannot due to this issue.

olimart:
Hi,
can you submit PR?

PR #402, opened 2018-01-15
Reduce db calls in #revisions method

fatkodima:
To select revisions for record, `#revisions` method performs a lot of unnecessary db calls.

1. selects all revisions from the beginning https://github.com/collectiveidea/audited/blob/0c4730c2c357211fd4e1e0396ee5ed5e0706f525/lib/audited/auditor.rb#L106
2. selects unnecessary (all in this case) fields https://github.com/collectiveidea/audited/blob/0c4730c2c357211fd4e1e0396ee5ed5e0706f525/lib/audited/audit.rb#L60

And for **every** revision it makes 2 additional queries:
1. fetches user record again and again https://github.com/collectiveidea/audited/blob/0c4730c2c357211fd4e1e0396ee5ed5e0706f525/lib/audited/audit.rb#L67
2. fetches ancestor revisions again and again https://github.com/collectiveidea/audited/blob/0c4730c2c357211fd4e1e0396ee5ed5e0706f525/lib/audited/audit.rb#L68

**Difference**:
```diff
   Audited::Audit Exists (0.8ms)  SELECT  1 AS one FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (version >= 1) LIMIT $3  [["auditable_id", 1], ["auditable_type", "User"], ["LIMIT", 1]]
-  Audited::Audit Load (0.4ms)  SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (version >= 1) ORDER BY "audits"."version" ASC  [["auditable_id", 1], ["auditable_type", "User"]]
-  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
-  Audited::Audit Load (0.4ms)  SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (version <= 1) ORDER BY "audits"."version" ASC  [["auditable_id", 1], ["auditable_type", "User"]]
-  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
-  Audited::Audit Load (0.4ms)  SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (version <= 2) ORDER BY "audits"."version" ASC  [["auditable_id", 1], ["auditable_type", "User"]]
-  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
-  Audited::Audit Load (0.3ms)  SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 AND (version <= 3) ORDER BY "audits"."version" ASC  [["auditable_id", 1], ["auditable_type", "User"]]
+  Audited::Audit Load (0.3ms)  SELECT "audits"."audited_changes", "audits"."version" FROM "audits" WHERE "audits"."auditable_id" = $1 AND "audits"."auditable_type" = $2 ORDER BY "audits"."version" ASC  [["auditable_id", 1], ["auditable_type", "User"]]
 => [#<User id: 1, name: "fatkodima", email: nil, age: 25, company_id: 1, version: 1>, #<User id: 1, name: "fatkod", email: nil, age: 25, company_id: 1, version: 2>, #<User id: 1, name: "fatkod", email: nil, age: 30, company_id: 1, version: 3>]
```

Also fixes https://github.com/collectiveidea/audited/issues/379.


tbrisker:
@fatkodima Looks good in general, but it seems to break Rails 4.0 - care to take a look?

lib/audited/auditor.rb
```diff
@@ -101,9 +101,15 @@ def without_auditing(&block)
       #   end
       #
       def revisions(from_version = 1)
-        audits = self.audits.from_version(from_version)
-        return [] if audits.empty?
-        audits.map(&:revision)
+        return [] if audits.from_version(from_version).empty?
+
+        loaded_audits = audits.select(:audited_changes, :version).to_a
```
tbrisker: You can use `audits.pluck(:audited_changes, :version)` here, I think that will work on all rails versions.

fatkodima: Yeah, I can, but just don't want to deal with raw arrays and complicate code a bit.

fatkodima:
@tbrisker Looks like working now.

tbrisker:
Thank you @fatkodima ! I think we can further improve the performance here - right now, this loops over the audits in memory multiple times and reconstructs the attribute hash over and over for every version, which can be costly when there is a large number of versions (O(n<sup>2</sup>) complexity), but that can be handled in a separate PR.

PR #403, opened 2018-01-15
Optimize reconstructing attributes in #revisions

fatkodima:
Follow up of https://github.com/collectiveidea/audited/pull/402.

Also replaced `.empty?` with `.exists?`, because for older rails'es (at least for 4.0.0) `.empty?` is implemented through `COUNT`, instead of through `SELECT 1` like in new versions.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -101,14 +101,16 @@ module Audited
       #   end
       #
       def revisions(from_version = 1)
-        return [] if audits.from_version(from_version).empty?
+        return [] unless audits.from_version(from_version).exists?
 
-        loaded_audits = audits.select([:audited_changes, :version]).to_a
-        targeted_audits = loaded_audits.select { |audit| audit.version >= from_version }
+        all_audits = audits.select([:audited_changes, :version]).to_a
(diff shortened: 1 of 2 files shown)


lib/audited/auditor.rb
```diff
@@ -101,14 +101,17 @@ def without_auditing(&block)
...
+        targeted_audits = all_audits.select { |audit| audit.version >= from_version }
+
+        previous_attributes = reconstruct_attributes(all_audits - targeted_audits)
 
         targeted_audits.map do |audit|
-          ancestors = loaded_audits.select { |a| a.version <= audit.version  }
-          revision_with(reconstruct_attributes(ancestors).merge(version: audit.version))
+          revision_attributes = previous_attributes.merge(audit.new_attributes)
+          previous_attributes = revision_attributes
+          revision_with(revision_attributes.merge!(version: audit.version))
```
tbrisker: I think you don't actually need the extra variable here, maybe something like:
```ruby
previous_attributes.merge!(audit.new_attributes)
revision_with(previous_attributes.merge!(version: audit.version))

fatkodima:
Updated this. Failing tests are not related to my changes. I fixed them in [a separate PR](https://github.com/collectiveidea/audited/pull/409).

Issue #274, opened 2016-05-02
Limit number of audits stored


jeremylynch:
It would be handy if there was a way to store only a certain number of audits. Ie. delete the first audit once this number if reached. 

I am currently achieving this with the following code. 

``` ruby
def update_attributes_with_audits(case_params)
  update(case_params)
  audits_count = audits.count
  if audits_count >= 15
    audits.first.destroy
  end
end
```

lacco:
I want to achieve something similar. I haven't looked into the exact details yet, but instead of just destroying old audits, I want to merge the changes in newer ones. I would move the logic in a daily cronjob and merge multiple ```update``` audits.

In case somebody has already a snippet for this, it would be helpful. I will also share my results here since it feels like that this might be related.

fatkodima:
Hey guys, i've made a pull request for that - #405

PR #405, opened 2018-01-16
Limit number of audits stored

fatkodima:
This is working by simply merging first two audits into second and deleting first, when keeping amount exceeds limit.

Probably might need to add the ability to allow set `:keep_audits` from initializer and add some documentation? 

Related to https://github.com/collectiveidea/audited/issues/274.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -105,10 +105,11 @@ module Audited
       end
     end
 
-    def merge!(other)
+    def merge(other)
       raise ArgumentError, "#{other.inspect} should belong to the same auditable" \
         if other.auditable_type != auditable_type || other.auditable_id != auditable_id
-      update_column(:audited_changes, other.audited_changes.merge(audited_changes))
+      audited_changes.reverse_merge!(other.audited_changes)
+      self
     end
 
     # Allows user to be set to either a string or an ActiveRecord object
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -105,13 +105,6 @@ module Audited
       end
     end
 
-    def merge(other)
-      raise ArgumentError, "#{other.inspect} should belong to the same auditable" \
-        if other.auditable_type != auditable_type || other.auditable_id != auditable_id
-      audited_changes.reverse_merge!(other.audited_changes)
-      self
-    end
-
     # Allows user to be set to either a string or an ActiveRecord object
     # @private
     def user_as_string=(user)
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -105,6 +105,12 @@ module Audited
       end
     end
 
+    def merge!(other)
+      raise ArgumentError, "#{other.inspect} should belong to the same auditable" \
+        if other.auditable_type != auditable_type || other.auditable_id != auditable_id
+      update_column(:audited_changes, other.audited_changes.merge(audited_changes))
+    end
+
     # Allows user to be set to either a string or an ActiveRecord object
     # @private
     def user_as_string=(user)
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -33,6 +33,7 @@ module Audited
       #
       # * +require_comment+ - Ensures that audit_comment is supplied before
       #   any create, update or destroy operation.
+      # * +max_audits+ - Limits the number of stored audits.
       #
       # * +if+ - Only audit the model when the given function returns true
       # * +unless+ - Only audit the model when the given function returns false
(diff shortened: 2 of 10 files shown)


lib/audited/audit.rb
```diff
@@ -105,6 +105,12 @@ def undo
       end
     end
 
+    def merge!(other)
+      raise ArgumentError, "#{other.inspect} should belong to the same auditable" \
+        if other.auditable_type != auditable_type || other.auditable_id != auditable_id
+      update_column(:audited_changes, other.audited_changes.merge(audited_changes))
```
tbrisker: I think this method should be named `reverse_merge!`, so it is clear which one has precedence in a consistent manner to common ruby `merge!` methods, or named something completely different.

lib/audited/auditor.rb
```diff
@@ -202,7 +204,13 @@ def audit_create
       def audit_update
         unless (changes = audited_changes).empty? && audit_comment.blank?
           write_audit(action: 'update', audited_changes: changes,
-                      comment: audit_comment)
+                      comment: audit_comment) do
+            if keep_audits && audits.count > keep_audits.to_i
+              first_audit, second_audit = audits.limit(2)
+              second_audit.merge!(first_audit)
```
tbrisker: what happens if the first audit is a create? could this cause issues?

lib/audited/auditor.rb
```diff
@@ -202,7 +204,13 @@ def audit_create
       def audit_update
         unless (changes = audited_changes).empty? && audit_comment.blank?
           write_audit(action: 'update', audited_changes: changes,
-                      comment: audit_comment)
+                      comment: audit_comment) do
+            if keep_audits && audits.count > keep_audits.to_i
```
tbrisker: Shouldn't this check be done inside `write_audit` function always so it runs on destroy audits as well? It would also reduce the complexity.

lib/audited/auditor.rb
```diff
@@ -43,12 +43,14 @@ def audited(options = {})
 
         class_attribute :audit_associated_with,   instance_writer: false
         class_attribute :audited_options,       instance_writer: false
+        class_attribute :keep_audits,             instance_writer: false
```
tbrisker: Perhaps `max_audits` is more descriptive of what this setting is? Maybe also ensure it is a positive integer in the option normalization?

lib/audited/audit.rb
fatkodima: Looks like it's clear that caller is taking precedence over `other` (like for common `merge` methods).
This line is same as
```
self.update_column(....
```
So we are updating self with contents of other.

fatkodima:
Updated this.

lib/audited/auditor.rb
```diff
@@ -216,7 +219,24 @@ def audit_destroy
...
+      end
+
+      def reduce_audits_if_needed
+        if max_audits && (extra_count = audits.count - max_audits) > 0
+          first, *extra, merge_target = audits.limit(extra_count + 1)
+          (extra + [merge_target]).inject(first) { |memo, audit| audit.merge(memo) }
+
+          merge_target.save!
+          audits.where(id: ([first] + extra).map(&:id)).delete_all
+        end
```
tbrisker: This function is quite complex as it is, perhaps it would be best to split into two? 
so this one only checks which audits if any need to be combined and sends them to a second function, say `combine_audits` that gets a relation and combines all the audits. that would be beneficial also so users can create rake tasks to bulk combine audits (which could be important when adding this to existing databases, as the first run may be very long) or to allow future capabilities (for example, collapse audits older then a certain date). 
Another possible optimization (which would possibly make the audit merge method unnecessary) would be to use something along the lines of `audits_to_combine.last.audited_changes = audits_to_combine.pluck(:audited_changes).reduce(&:merge)`.
One last comment, when collapsing audits it might be a good idea to add a line to the audit comment indicating this audit is the result of multiple audits being combined or indicate this in some other way.

lib/audited/auditor.rb
```diff
@@ -43,12 +44,14 @@ def audited(options = {})
...
         class_attribute :audit_associated_with,   instance_writer: false
         class_attribute :audited_options,       instance_writer: false
+        class_attribute :max_audits,              instance_writer: false
         attr_accessor :version, :audit_comment
 
         self.audited_options = options
         normalize_audited_options
 
         self.audit_associated_with = audited_options[:associated_with]
+        self.max_audits = audited_options[:max_audits] || Audited.max_audits
```
tbrisker: Shouldn't the fallback to default be done in the options normalization?

spec/rails_app/config/initializers/audited.rb
```diff
@@ -0,0 +1,5 @@
+require 'audited'
+
+Audited.config do |config|
+  config.max_audits = 10
```
tbrisker: This might lead to unexpected results in future tests, perhaps only mock this for tests related to this option?

fatkodima:
@tbrisker Sorry for delay. Updated with your requests

tbrisker:
Thanks for the effort @fatkodima !

fatkodima:
🎉 😄

nynhex:
@fatkodima Thanks!! <3 I like the option of a limit in the initializer.

thejasbabu:
This somehow does not seem to work for me
Audited version  (4.8.0)
Rails version (5.2.2)

I have initialised with `Audited.max_audits = 5` but still find more than 5 audit present in the db.

thejasbabu:
I think this does not work for associated audits. 
For eg for user belonging to company: 

`audited associated_with: :company, max_audits: 2`

This should work?

gathuku:
Is it possible to define max_audits in a proc?
```rb
audited max_audits: -> {  method_to_find_max_audits }
```

PR #409, opened 2018-01-31
Fix migrations in tests for rails 5.2

fatkodima:
Rails 5.2.rc1 [introduced](https://github.com/rails/rails/pull/31727) some breaking changes for `ActiveRecord::Migrator`, so this PR fixes tests according to new changes.

### DIFF
--- a/spec/audited_spec_helpers.rb
+++ b/spec/audited_spec_helpers.rb
@@ -17,4 +17,16 @@ module AuditedSpecHelpers
     end
   end
 
+  def run_migrations(direction, migrations_paths, target_version = nil)
+    if rails_below?('5.2.0.rc1')
+      ActiveRecord::Migrator.send(direction, migrations_paths, target_version)
+    else
+      ActiveRecord::MigrationContext.new(migrations_paths).send(direction, target_version)
+    end
+  end
+
+  def rails_below?(rails_version)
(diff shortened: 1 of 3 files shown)


spec/audited_spec_helpers.rb
```diff
@@ -17,4 +17,12 @@ def create_versions(n = 2)
     end
   end
 
+  def run_migrations(direction, migrations_paths, target_version = nil)
+    if Rails.version >= '5.2.0.0'
```
tbrisker: This will fail once 5.2.0 is out, since `'5.2.0' >= '5.2.0.0' => false`.

fatkodima: ouch, extra zero.
Such `Rails.version >= '5.2.0'` will not work for `.beta2`, after which change was introduced. Should I bother of exactly greater than `.beta2` or is `Rails.version >= '5.2.0'` enough?

tbrisker: perhaps `Rails.version.start_with? '5.2'` for now, and change it later if needed for 6?

fatkodima:
Updated this to use exact version:
- to allow working on all 5.2 betas
- and to avoid updating code for new major rails releases as proposed in
> perhaps Rails.version.start_with? '5.2' for now, and change it later if needed for 6?

PR #167, opened 2014-02-06
Add conditional and exclusionary auditing feature

codenamev:
Add conditional and exclusionary auditing feature
Works similar to conditional callbacks

PR #414, opened 2018-02-22
Add conditional and exclusionary auditing feature

### LINKED ISSUE #237: Conditional auditing possible?
Awesome gem, thanks for creating this.

We have a model that handles both normal data and data that acts as a cache for performant querying. We only want to audit the former.

Is it possible to do something like this, similar to conditions used for callbacks:

``` ruby
class Availability < ActiveRecord::Base
  audited, if: Proc.new { |availability| availability.time_off? || availability.recurring? }
end
```

codenamev:
Works similar to conditional callbacks
* Adds conditional (:if) and exclusionary (:unless) auditing feature
* Updates docs for `audited` to reflect new features

Replaces #167 
Closes #237

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -52,8 +52,8 @@ module Audited
         extend Audited::Auditor::AuditedClassMethods
         include Audited::Auditor::AuditedInstanceMethods
 
-        class_attribute :audit_associated_with,      instance_writer: false
-        class_attribute :audited_options,            instance_writer: false
+        class_attribute :audit_associated_with, instance_writer: false
+        class_attribute :audited_options,       instance_writer: false
         attr_accessor :version, :audit_comment
 
         self.audited_options = options
@@ -256,7 +256,7 @@ module Audited
         return true if condition.respond_to?(:blank?) && condition.blank?
         return true if !condition.respond_to?(:call) and !respond_to?(condition.to_sym)
 
-        return condition.call == matching if condition.respond_to?(:call)
+        return condition.call(self) == matching if condition.respond_to?(:call)
         return send(condition) == matching if respond_to?(condition)
 
         true
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -245,7 +245,6 @@ module Audited
       def auditing_enabled
         if_condition = audited_options[:if]
         unless_condition = audited_options[:unless]
-        return self.class.auditing_enabled unless if_condition.present? || unless_condition.present?
 
         return run_conditional_check(audited_options[:if]) &&
           run_conditional_check(audited_options[:unless], matching: false) &&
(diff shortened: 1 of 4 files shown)


westonganger:
+1

westonganger:
This should be updated to allow Procs and lambdas

codenamev:
@tbrisker Updated with your requests, and also added inline procs for fun 😃

lib/audited/auditor.rb
```diff
@@ -232,7 +243,23 @@ def require_comment
       end
 
       def auditing_enabled
-        self.class.auditing_enabled
+        if_condition = audited_options[:if]
+        unless_condition = audited_options[:unless]
+        return self.class.auditing_enabled unless if_condition.present? || unless_condition.present?
```
tbrisker: This check is redundant, since the first check in `run_conditional_check` returns true if the condition is not present.

lib/audited/auditor.rb
```diff
@@ -232,7 +243,23 @@ def require_comment
...
+
+        return run_conditional_check(audited_options[:if]) &&
+          run_conditional_check(audited_options[:unless], matching: false) &&
+          self.class.auditing_enabled
+      end
+
+      def run_conditional_check(condition, matching: true)
+        return true if condition.respond_to?(:blank?) && condition.blank?
```
tbrisker: `blank?` is defined by activesupport on `Object`, so no need to check if the condition responds to it first.

lib/audited/auditor.rb
```diff
@@ -232,7 +243,23 @@ def require_comment
...
+        return true if !condition.respond_to?(:call) and !respond_to?(condition.to_sym)
```
tbrisker: This check is also redundant, if these are both false then the next two lines will not execute and the default `true` will be returned. Note also that `and` should not be used for boolean conditions, only for flow control.

codenamev:
@tbrisker Updated 😄 Thanks for the thorough review 👍

lib/audited/auditor.rb
```diff
@@ -232,7 +243,21 @@ def require_comment
       end
 
       def auditing_enabled
-        self.class.auditing_enabled
+        if_condition = audited_options[:if]
+        unless_condition = audited_options[:unless]
```
tbrisker: These two assignments can be removed as well

tbrisker:
@westonganger mind adding a line to CHANGELOG.md about this change? Other than that this should be good to merge.

codenamev:
@tbrisker Updated CHANGELOG 😄

westonganger:
I have one comment, its best not to use blank? in the gems code because that is an external dependency that does not need to be there. Instead of `blank?` use the following ruby only version:

```ruby
condition.empty? || condition.nil?
```

codenamev:
@westonganger This gem has AR as a dependency (which has a [dependency on activesupport](https://github.com/rails/rails/blob/master/activerecord/activerecord.gemspec#L31)).  With activesupport included, we get `blank?` for free.

tbrisker:
squashed and merged as  c1aa93b, Thanks a lot @codenamev!

PR #419, opened 2018-03-02
Respect the 'on' option when used with 'comment_required' option

jeffdill2:
This will prevent validations from requiring the `audit_comment` attribute to be present (when you're utilizing the `comment_required: true` option) even if you're not actually auditing the specific action that's taking place.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -51,7 +51,7 @@ module Audited
         self.audit_associated_with = audited_options[:associated_with]
 
         if audited_options[:comment_required]
-          validates_presence_of :audit_comment, if: :auditing_enabled
+          validate :presence_of_audit_comment
           before_destroy :require_comment
         end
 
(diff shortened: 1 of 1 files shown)


lib/audited/auditor.rb
```diff
@@ -219,11 +219,30 @@ def write_audit(attrs)
         run_callbacks(:audit)  { audits.create(attrs) } if auditing_enabled
       end
 
+      def presence_of_audit_comment
+        case
+        when !auditing_enabled
+          return true
+        when audit_comment.present?
+          return true
+        when audited_options[:on].present? && audited_options[:on].exclude?(:create) && self.new_record?
```
tbrisker: no need to check `audited_options[:on].present?` here and below, the `normalize_audited_options` function will make sure it's there and contains at least one option.

lib/audited/auditor.rb
```diff
@@ -51,7 +51,7 @@ def audited(options = {})
         self.audit_associated_with = audited_options[:associated_with]
 
         if audited_options[:comment_required]
-          validates_presence_of :audit_comment, if: :auditing_enabled
+          validate :presence_of_audit_comment
           before_destroy :require_comment
```
tbrisker: How about adding a condition to the callback here to `if audited_options[:on].include?(:destroy)` instead of the added check inside the function?

lib/audited/auditor.rb
jeffdill2: Ah, good call. I wasn't aware of the `normalize_audited_options` function. Thanks!

lib/audited/auditor.rb
jeffdill2: The original logic behind doing it the way I did is because I didn't want this to be skipped via `if audited_options[:on].include?(:destroy)` if you weren't utilizing the `on:` option at all (i.e. I was assuming `on:` would be empty if you weren't utilizing it), but now that I'm aware of the `normalize_audited_options` function, definitely makes more sense to have the condition directly on the callback. 😄

jeffdill2:
@tbrisker sure thing, I'll try to get some tests added today. 👍

jeffdill2:
@tbrisker should be good to go now.

tbrisker:
Thanks @jeffdill2 ! One last thing, sorry I forgot to mention before - could you please also add a line for this change to the CHANGELOG.md?

jeffdill2:
@tbrisker not a problem. CHANGELOG is updated now. 👍

tbrisker:
Thank you very much @jeffdill2!

jeffdill2:
@tbrisker my pleasure! 😀

PR #420, opened 2018-03-11
Fix RSpec matchers along with a few other things.

hernanat:
This commit fixes issues with the RSpec matchers, notably:

- For the `except` matcher code path, `model_class.default_ignored_attributes` was being called, but that method was made protected in a previous change as discussed [here](https://github.com/collectiveidea/audited/commit/8f8495dabdac0e44bdd97c5294177f3fdfc4caae#commitcomment-28001384). This commit makes that method public again.

- Both the `on` and `except` matcher options have undergone a pretty substantial refactor. As I was fixing the original issue with `on` and `except`, I started to notice other things and addressed them. I will try to make explanatory comments on the PR to give some insight.

- The matchers as a whole have undergone pretty significant changes, and I found that many of them were not working as intended, or, for example, in the case of `on`, it was a no-op. I made the changes that I thought reflected what the matchers ought to do and I also included tests (which actually ended up being extremely useful in discovering some of the bugs!) to make sure that future changes don't break anything.

- The `requires_comment` matcher's behavior has changed to look for callbacks instead. I also changed the way that comment-required state validity is checked in the `Auditor`, but I will discuss that below. My reasoning for the two changes are related though - given an audited model where audit comments are required, we only care about the existence of the comment when an object is being created, if it has been changed (and subsequently updated), or if it is about to be destroyed (and of course we can include & exclude any combination of these three states with the `on` option). Previously, the presence of the comment was related to the overall state of the object itself, and I found this confusing since the comment only matters when we are _doing_ something to the object.

- I have decoupled the codepaths for each option so that they could be run independently of one another (and all together, of course). I was getting some interesting behavior before, but I believe I have nailed down the correct behavior.

- I will discuss anything not discussed above for the matchers in PR comments.

The `Auditor` itself:

- Changed the way non-audited columns were calculated. Will discuss in-line.

- Revised and (hopefully) simplified the way that audit comments are required & validated. Will discuss in-line.

Other comments:

- As I mentioned above, I noticed some behavioral issues that I tried to address. I will talk about them more in-line and if there are any questions about particular things I did please feel free to ask.

- Another thing I would like to do is a complete rework of the test suite itself (or if anyone else wants to take this up, feel free :P) In writing the tests for the matchers, I noticed many things such as audited columns for the `User` model being inconsistent, the `Audited.ignored_attributes` changing, and more. This is directly caused by the behavior of some of the tests, and I think it would be beneficial to figure out a way to rework the behavior such that what happens in one test doesn't have an impact on all the others. I also have a slight fear that some tests may in fact be returning false-positives, but I haven't looked too far into that yet.

- I was wondering if it would be better to not memoize `Auditor#audited_columns` and `Auditor#non_audited_columns` since the calculations are not really intensive enough to warrant memoization, but when I did that tests were breaking left and right and so I think this really ties into my point above about the test suite. Something to keep in mind / look into. I didn't do it in this PR because there is already quite a bit going on.

### DIFF
--- a/gemfiles/rails52.gemfile
+++ b/gemfiles/rails52.gemfile
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "rails", ">= 5.2.0.rc1", "< 5.3"
+gem "rails", ">= 5.2.0.beta2", "< 5.3"
 gem "mysql2", "~> 0.4.4"
 
 gemspec name: "audited", path: "../"
--- a/gemfiles/rails52.gemfile
+++ b/gemfiles/rails52.gemfile
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "rails", ">= 5.2.0.beta2", "< 5.3"
+gem "rails", ">= 5.2.0.rc1", "< 5.3"
 gem "mysql2", "~> 0.4.4"
 
 gemspec name: "audited", path: "../"
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -64,6 +64,8 @@ module Audited
 
         if audited_options[:comment_required]
           validate :presence_of_audit_comment
+          before_create :require_comment if audited_options[:on].include?(:create)
+          before_update :require_comment if audited_options[:on].include?(:update)
           before_destroy :require_comment if audited_options[:on].include?(:destroy)
         end
 
@@ -251,20 +253,7 @@ module Audited
       end
 
       def presence_of_audit_comment
-        case
-        when !auditing_enabled
-          return true
-        when audit_comment.present?
-          return true
-        when audited_options[:on].exclude?(:create) && self.new_record?
-          return true
-        when audited_options[:on].exclude?(:update) && self.persisted?
-          return true
-        when audited_changes.empty? && self.persisted?
-          return true
-        else
-          errors.add(:audit_comment, "can't be blank.")
-        end
+        require_comment if comment_required_state?
       end
 
       def combine_audits_if_needed
@@ -277,10 +266,15 @@ module Audited
 
       def require_comment
         if auditing_enabled && audit_comment.blank?
-          errors.add(:audit_comment, "Comment required before destruction")
+          errors.add(:audit_comment, "Comment can't be blank!")
           return false if Rails.version.start_with?('4.')
-          throw :abort
+          throw(:abort)
         end
+      end
+
+      def comment_required_state?
+        (audited_options[:on].include?(:create) && self.new_record?) ||
+          (audited_options[:on].include?(:update) && self.persisted? && self.changed?)
       end
 
       CALLBACKS.each do |attr_name|
@@ -321,9 +315,7 @@ module Audited
 
       # We have to calculate this here since column_names may not be available when `audited` is called
       def non_audited_columns
-        @non_audited_columns ||= audited_options[:only].present? ?
-                                 column_names - audited_options[:only] :
-                                 default_ignored_attributes | audited_options[:except]
+        @non_audited_columns ||= calculate_non_audited_columns
       end
 
       def non_audited_columns=(columns)
@@ -369,10 +361,11 @@ module Audited
         Audited.store["#{table_name}_auditing_enabled"] = val
       end
 
-      protected
       def default_ignored_attributes
-        [primary_key, inheritance_column] + Audited.ignored_attributes
+        [primary_key, inheritance_column] | Audited.ignored_attributes
       end
+
+      protected
 
       def normalize_audited_options
         audited_options[:on] = Array.wrap(audited_options[:on])
@@ -381,6 +374,16 @@ module Audited
         audited_options[:except] = Array.wrap(audited_options[:except]).map(&:to_s)
         max_audits = audited_options[:max_audits] || Audited.max_audits
         audited_options[:max_audits] = Integer(max_audits).abs if max_audits
+      end
+
+      def calculate_non_audited_columns
+        if audited_options[:only].present?
+          (column_names | default_ignored_attributes) - audited_options[:only]
+        elsif audited_options[:except].present?
+          default_ignored_attributes | audited_options[:except]
+        else
+          default_ignored_attributes
+        end
       end
     end
   end
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -64,8 +64,6 @@ module Audited
 
         if audited_options[:comment_required]
           validate :presence_of_audit_comment
-          before_create :require_comment if audited_options[:on].include?(:create)
-          before_update :require_comment if audited_options[:on].include?(:update)
           before_destroy :require_comment if audited_options[:on].include?(:destroy)
         end
 
@@ -253,7 +251,15 @@ module Audited
       end
 
       def presence_of_audit_comment
-        require_comment if comment_required_state?
+        if comment_required_state?
+          errors.add(:audit_comment, "Comment can't be blank!") unless audit_comment.present?
+        end
+      end
+
+      def comment_required_state?
+        auditing_enabled &&
+          ((audited_options[:on].include?(:create) && self.new_record?) ||
+          (audited_options[:on].include?(:update) && self.persisted? && self.changed?))
       end
 
       def combine_audits_if_needed
@@ -270,11 +276,6 @@ module Audited
           return false if Rails.version.start_with?('4.')
           throw(:abort)
         end
-      end
-
-      def comment_required_state?
-        (audited_options[:on].include?(:create) && self.new_record?) ||
-          (audited_options[:on].include?(:update) && self.persisted? && self.changed?)
       end
 
       CALLBACKS.each do |attr_name|
--- a/lib/audited/rspec_matchers.rb
+++ b/lib/audited/rspec_matchers.rb
@@ -114,7 +114,7 @@ module Audited
 
       def comment_required_valid?
         expects "to require audit_comment before #{model_class.audited_options[:on]} when comment required"
-        requires_comment_before_callbacks?
+        validate_callbacks_include_presence_of_comment? && destroy_callbacks_include_comment_required?
       end
 
       def only_audit_on_designated_callbacks?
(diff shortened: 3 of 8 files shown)


gemfiles/rails52.gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "rails", ">= 5.2.0.rc1", "< 5.3"
+gem "rails", ">= 5.2.0.beta2", "< 5.3"
```
hernanat: This was automatically generated when I ran appraisal.

lib/audited/auditor.rb
```diff
@@ -52,6 +52,8 @@ def audited(options = {})
 
         if audited_options[:comment_required]
           validate :presence_of_audit_comment
+          before_create :require_comment if audited_options[:on].include?(:create)
```
hernanat: so before we were trying to cram all of this comment requiring behavior just into the state of the object, and then for destroy we were using a callback. I think it make sense to use a callback for all three, and to have a more natural way of validating presence of an audit comment before any of these are called, which I will discuss below.

lib/audited/auditor.rb
```diff
@@ -220,30 +222,21 @@ def write_audit(attrs)
       end
 
       def presence_of_audit_comment
-        case
-        when !auditing_enabled
```
hernanat: This method was misleading, and again, I think it was trying to do too much / wasn't doing what it said it was.

lib/audited/auditor.rb
```diff
@@ -220,30 +222,21 @@ def write_audit(attrs)
...
-        when audited_changes.empty? && self.persisted?
-          return true
-        else
-          errors.add(:audit_comment, "can't be blank.")
-        end
+        require_comment if comment_required_state?
       end
 
       def require_comment
         if auditing_enabled && audit_comment.blank?
```
hernanat: So, in the case of callbacks, if this is getting called by `before_create`, `before_update`, etc. then all we need to check are these two things. If these criteria aren't met, then the state is invalid and it won't be created/updated/destroyed.

lib/audited/auditor.rb
```diff
@@ -220,30 +222,21 @@ def write_audit(attrs)
...
-          errors.add(:audit_comment, "Comment required before destruction")
-          return false if Rails.version.start_with?('4.')
-          throw :abort
+          errors.add(:audit_comment, "Comment can't be blank!")
+          throw(:abort)
         end
       end
 
+      def comment_required_state?
+        (audited_options[:on].include?(:create) && self.new_record?) ||
```
hernanat: Right. So, here we are with what I was talking about the validity of the object while in a given state, but not being created / updated. My reasoning is that we only care about the existence of the comment when:

1.) The object is a new record and we require a comment + audit on create

2.) The object is an existing record, has had some attributes changed, and we require a comment + audit on update.

lib/audited/auditor.rb
```diff
@@ -271,9 +264,7 @@ def audited_columns
 
       # We have to calculate this here since column_names may not be available when `audited` is called
       def non_audited_columns
-        @non_audited_columns ||= audited_options[:only].present? ?
-                                 column_names - audited_options[:only] :
-                                 default_ignored_attributes | audited_options[:except]
+        @non_audited_columns ||= calculate_non_audited_columns
```
hernanat: For here and also `audited_columns`, I am not sure that we gain anything from memoizing these values. And as I mentioned in my commit message, taking away the memoization breaks a bunch of stuff which I think may hint at the existence of another problem. Not sure though.

lib/audited/auditor.rb
```diff
@@ -319,17 +310,28 @@ def auditing_enabled=(val)
         Audited.store["#{table_name}_auditing_enabled"] = val
       end
 
-      protected
       def default_ignored_attributes
-        [primary_key, inheritance_column] + Audited.ignored_attributes
+        [primary_key, inheritance_column] | Audited.ignored_attributes
```
hernanat: I changed this to a `|` because if `Audited.ignored_attributes` because it is not inconceivable that `+` could lead to duplicate values say if someone configured `Audited.ignored_attributes` to include `inheritance_column`.

lib/audited/auditor.rb
```diff
@@ -319,17 +310,28 @@ def auditing_enabled=(val)
...
+      protected
+
       def normalize_audited_options
         audited_options[:on] = Array.wrap(audited_options[:on])
         audited_options[:on] = [:create, :update, :destroy] if audited_options[:on].empty?
         audited_options[:only] = Array.wrap(audited_options[:only]).map(&:to_s)
         audited_options[:except] = Array.wrap(audited_options[:except]).map(&:to_s)
       end
+
+      def calculate_non_audited_columns
```
hernanat: So I changed this calculation up a bit. Mostly to drive home the point that the `default_ignored_attributes` should always be included. Previously, they were only considered for the `except` code path and so `only` vs `except` had inconsistent construction logic for deciding which columns to ignore. Here are some other notes I wrote to myself in comments on this: 

```rb
# since there are technically three states
# i.e. audit _only_ these columns, or _except_ these columns
# or _all_ columns (except default ignored), we should probably be explicit about this
# the previously written code was not and it was pretty hard to follow
# and track down bugs. This proposed "new" way has the benefit of being
# more explicit and thus the intent is more obvious
# also the way it was written before broke the rspec matchers since
# in the previous way there would be extra columns include for
# the `except` option because of the original `default_ignored_attributes`
```

lib/audited/rspec_matchers.rb
```diff
@@ -109,31 +106,63 @@ def associated_with_model?
...
+        end.compact.all?
+      end
+
+      def callbacks_for(action, kind: :before)
+        model_class.send("_#{action}_callbacks").select { |cb| cb.kind == kind }.map(&:filter)
+      end
 
-          expects "to be invalid when audit_comment is not specified"
-          @subject.valid? == false && @subject.errors.key?(:audit_comment)
+      def build_ignored_fields_from_options
```
hernanat: This logic originally lived in `records_changes_to_specified_fields?`. I wrote some comments to myself on that method and this one which I share below:


```rb
# second branch of the case statement was conditionally unioning `except`
# if the `except` option was present, otherwise it would just return default
# ignored fields; while logically correct, I thought it might be clearer
# if we were explicit about the cases in wich the ignored attributes were build
# i.e. either when `only` is present, `except` is present, or neither is present
# I also changed `except` to `ignored_fields` because I found it a bit confusing
# to be using a variable name that matched one of the options to the auditor
# even though the values within the array weren't necessarily the values
# passed into the option array

# I also got rid of the default `true` because even if neither option is passed in
# we should probably check and see that the default ignored fields are still
# not being audited
```

lib/audited/rspec_matchers.rb
```diff
@@ -109,31 +106,63 @@ def associated_with_model?
...
       end
 
       def comment_required_valid?
-        if @options[:comment_required]
-          @subject.audit_comment = nil
+        expects "to require audit_comment before #{model_class.audited_options[:on]} when comment required"
+        requires_comment_before_callbacks?
+      end
+
+      def only_audit_on_designated_callbacks?
```
hernanat: yeah...I went meta...lol.

lib/audited/rspec_matchers.rb
```diff
@@ -109,31 +106,63 @@ def associated_with_model?
...
+          (default_ignored_attributes | model_class.column_names) - @options[:only]
+        elsif @options[:except].present?
+          default_ignored_attributes | @options[:except]
         else
-          true
+          default_ignored_attributes
         end
       end
+
+      def required_checks_for_options_satisfied?
```
hernanat: I did this to ensure that only the checks for options that are included are being ran. This helped get rid of all the default `true`.

spec/audited/auditor_spec.rb
```diff
@@ -545,21 +545,21 @@ def non_column_attr=(val)
 
     describe "on create" do
       it "should not validate when audit_comment is not supplied" do
-        expect(Models::ActiveRecord::CommentRequiredUser.new).not_to be_valid
```
hernanat: hmm...could probably still use this case actually.

spec/audited/rspec_matchers_spec.rb
```diff
@@ -0,0 +1,62 @@
+require "spec_helper"
+
+describe Models::ActiveRecord::UserExceptPassword do
+  let(:non_audited_columns) { subject.class.non_audited_columns }
+
+  it { should_not be_audited.only(non_audited_columns) }
```
hernanat: I'm trying to drive home the point that these should work independently of each other. Might be helpful to have a case chaining them together as well.

spec/audited_spec_helpers.rb
```diff
@@ -1,5 +1,13 @@
 module AuditedSpecHelpers
 
+  def reset_audited_ignored_attributes
+    Audited.ignored_attributes = %w(lock_version created_at updated_at created_on updated_on)
+  end
+
+  def reset_user_non_audited_fields
```
hernanat: woops. These are leftover from me trying to circumvent issues that arose within the test suite which I mentioned in my commit message.

spec/support/active_record/models.rb
```diff
@@ -13,6 +13,11 @@ def name=(val)
       end
     end
 
+    class UserExceptPassword < ::ActiveRecord::Base
```
hernanat: I included this to get around using the `User` model because of the issues I mentioned with respect to the test suite changing the behavior of things.

hernanat:
cc @darryn02 @tbrisker @margaretoti

gemfiles/rails52.gemfile
hernanat: Ahhh. Should I change it back?

tbrisker: Looks like #409 missed updating the Appraisal file, fixed in 402c31ecd792ac52294bf1744660d7c72bea22fb

hernanat:
@tbrisker I think I addressed all the comments. And yeah I agree, in the future I will definitely chunk it out into smaller PRs. Thanks for review / advice :)

tbrisker:
Thanks for the great work @hernanat! :trophy:

PR #356, opened 2017-06-18
add actionable coverage reporting

grosser:
minimal overhead and instant/actionable feedback when running tests ... found 1 duplicate method and a seemingly dangerous setter ... doing `foo.auditing_enabled = false` should not change the class, just trouble waiting to happen ...

```
rspec spec/audited/auditor_spec.rb
.................................................................................

Finished in 1.79 seconds (files took 2.81 seconds to load)
81 examples, 0 failures

Randomized with seed 2971

lib/audited/auditor.rb new uncovered lines introduced (2 current vs 0 configured)
Lines missing coverage:
lib/audited/auditor.rb:161
lib/audited/auditor.rb:243
```

PR #424, opened 2018-03-17
remove unused/untested methods

grosser:
auditing_enabled was especially dangerous since it sets on the instance and modifies the global state

@tbrisker broken out from https://github.com/collectiveidea/audited/pull/356

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -142,7 +142,7 @@ module Audited
 
       # List of attributes that are audited.
       def audited_attributes
-        attributes.except(*non_audited_columns)
+        attributes.except(*self.class.non_audited_columns)
       end
 
       # Combine multiple audits into one.
@@ -158,14 +158,6 @@ module Audited
       end
 
       protected
-
-      def non_audited_columns
-        self.class.non_audited_columns
-      end
-
-      def audited_columns
-        self.class.audited_columns
-      end
 
       def revision_with(attributes)
         dup.tap do |revision|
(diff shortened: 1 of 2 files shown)


grosser:
@tbrisker I can haz review plz

lib/audited/auditor.rb
```diff
@@ -204,7 +200,7 @@ def audited_changes
         if audited_options[:only].present?
           all_changes.slice(*audited_columns)
         else
-          all_changes.except(*non_audited_columns)
+          all_changes.except(*self.class.non_audited_columns)
```
tbrisker: Why is this change needed? why does it matter if we have an instance method that calls the class method just for convenience?

lib/audited/auditor.rb
```diff
@@ -297,10 +293,6 @@ def run_conditional_check(condition, matching: true)
         true
       end
 
-      def auditing_enabled=(val)
-        self.class.auditing_enabled = val
-      end
```
tbrisker: I'm OK with removing this method, but please add a line in the CHANGELOG file indicating this change and that the class method should be used instead.

lib/audited/auditor.rb
grosser: - having an instance method for a class variable makes it look like it's changeable/configured on the instance, so extra mental overhead and some users might think it's ok to do `instance.non_audited_columns << :foo`
 - it's not much convenience if it's only needed once
 - for symmetry also removed `def audited_columns `

lib/audited/auditor.rb
grosser: done

PR #425, opened 2018-03-20
Re-fix honoring of 'except' when used with 'comment_required'

jeffdill2:
This was originally fixed here: https://github.com/collectiveidea/audited/pull/419

But it was inadvertently broken when this was merged in: https://github.com/collectiveidea/audited/pull/420

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -63,7 +63,7 @@ module Audited
         self.audit_associated_with = audited_options[:associated_with]
 
         if audited_options[:comment_required]
-          validate :presence_of_audit_comment
+          validates_presence_of :audit_comment, if: :eligible_for_comment_validation?
           before_destroy :require_comment if audited_options[:on].include?(:destroy)
         end
 
@@ -250,16 +250,16 @@ module Audited
         end
       end
 
-      def presence_of_audit_comment
-        if comment_required_state?
-          errors.add(:audit_comment, "Comment can't be blank!") unless audit_comment.present?
+      def eligible_for_comment_validation?
+        if !auditing_enabled ||
+          (audited_options[:on].exclude?(:create) && self.new_record?) ||
+          (audited_options[:on].exclude?(:update) && self.persisted?) ||
+          (audited_changes.empty? && self.persisted?)
+        then
+          false
+        else
+          true
         end
-      end
-
-      def comment_required_state?
-        auditing_enabled &&
-          ((audited_options[:on].include?(:create) && self.new_record?) ||
-          (audited_options[:on].include?(:update) && self.persisted? && self.changed?))
       end
 
       def combine_audits_if_needed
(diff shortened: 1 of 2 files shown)


lib/audited/auditor.rb
```diff
@@ -250,18 +250,18 @@ def write_audit(attrs)
...
+          (audited_changes.empty? && self.persisted?)
+        then
+          false
+        else
+          true
         end
       end
 
-      def comment_required_state?
-        auditing_enabled &&
```
tbrisker: Wouldn't it be simpler to add here `audited_changes.peresent?`

tbrisker:
ping @jeffdill2, any updates on this?

jeffdill2:
@tbrisker OK, I've got a test in place now to ensure this doesn't break again in the future. 😄

jeffdill2:
@tbrisker Just following up with this one as well. Tests are in place now, so it's ready for review. Thanks!

lib/audited/auditor.rb
```diff
@@ -250,18 +250,18 @@ def write_audit(attrs)
...
-          errors.add(:audit_comment, "Comment can't be blank!") unless audit_comment.present?
+      def eligible_for_comment_validation?
+        if !auditing_enabled ||
+          (audited_options[:on].exclude?(:create) && self.new_record?) ||
+          (audited_options[:on].exclude?(:update) && self.persisted?) ||
+          (audited_changes.empty? && self.persisted?)
+        then
+          false
+        else
+          true
```
tbrisker: This is quite a confusing check, returning the opposite result to the condition - perhaps you can just reuse the `comment_required_state?` function and add the additional condition instead of flipping the entire logic around?

jeffdill2: I think it was mainly just the name `comment_required_state?` that I wasn't a fan of. To me it seems to indicate whether or not `comment_required` is enabled, but what it's really doing is adding additional checks for whether or not `audit_comment` should be validated for presence when `comment_required` is enabled. I felt that `eligible_for_comment_validation?` was clearer about what it's actually checking.

And my though process for flipping the logic around a bit was just so that the conditions would fail earlier – meaning that if auditing isn't enabled, we can immediately return `false` instead of having to test any more conditions. But I can definitely understand how it's a bit confusing. And now that I think about it, my logic was complete non-sense anyways – `if !auditing_enabled ||` and `auditing_enabled &&` will both return false immediately if auditing isn't enabled. So that was just a big derp on my part. 😬 

I'll get this fixed today.

jeffdill2:
@tbrisker OK, should be better now. 😄

jeffdill2:
@tbrisker merge conflict resolved now.

jeffdill2:
@tbrisker I've got `master` merged in now so checks are passing. This should be ready for review/merge whenever you get a chance. 😄

PR #426, opened 2018-03-24
Add ability to globally disable auditing

### LINKED ISSUE #410: Best way to disable auditing in test environment?
Hi, I'm currently working on a medium-sized Rails 4 project where Audited is being used extensively. The test suite relies pretty heavily on capybara features, so the performance overhead audited is adding to our test suite is really noticeable and I'm looking for a way to get rid of it.

I've tried different approaches to disable Audited only in test environment; however the only one that is working is putting

```
  audited unless Rails.env.test?
```

in every audited model. This has downsides - first of all it's really verbose; secondly, being this evaluated just once (at class evaluation time), it doesn't allow turning audited back on if I want to test audits in some specific tests.

I've also tried putting this in my `before(:suite)` method:

```
    Audited.audit_class.audited_class_names.map(&:constantize).each do |klass|
      klass.auditing_enabled = false
    end
```
and turning audit back on in individual specs like:

```
  around :each do |example|
    audit_enabled = User.auditing_enabled
    User.auditing_enabled = true
    example.run
    User.auditing_enabled = audit_enabled
  end
```

but it seems it's much slower than the solution above (25% slower on the overall suite running time).

[EDIT]: the code above it's slower cause `Audited.audit_class.audited_class_names == []` in the `before(:suite)`, as obviously the environment hasn't still been loaded, therefor making the lines above completely useless. It works by putting it in a `before(:all)` block, but I don't like having to repeat that code for every example group in my suite.

What's the suggested solution for this problem? Thanks.

fatkodima:
For minitest, for example:
Add to `test/test_helper.rb`:
```ruby
Audited.disable_auditing
```
And use:
```ruby
class UserTest < Minitest::Test
  def setup
    Audited.enable_auditing
  end

  def teardown
    Audited.disable_auditing
  end

  def test_auditing
    user = User.create!(name: "Dima", username: "dima")
    assert_equal 0, user.audits.count 
  end
end
```

Fixes #410

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -9,6 +9,18 @@ module Audited
       @audit_class ||= Audit
     end
 
+    def enable_auditing
+      store[:auditing_enabled] = nil
+    end
+
+    def disable_auditing
+      store[:auditing_enabled] = false
+    end
+
+    def auditing_enabled?
+      store[:auditing_enabled].nil? || store[:auditing_enabled]
+    end
+
     def store
       Thread.current[:audited_store] ||= {}
     end
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -2,23 +2,11 @@ require 'active_record'
 
 module Audited
   class << self
-    attr_accessor :ignored_attributes, :current_user_method, :max_audits
+    attr_accessor :ignored_attributes, :current_user_method, :max_audits, :auditing_enabled
     attr_writer :audit_class
 
     def audit_class
       @audit_class ||= Audit
-    end
-
-    def enable_auditing
-      store[:auditing_enabled] = nil
-    end
-
-    def disable_auditing
-      store[:auditing_enabled] = false
-    end
-
-    def auditing_enabled?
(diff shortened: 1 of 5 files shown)


lib/audited.rb
```diff
@@ -9,6 +9,18 @@ def audit_class
       @audit_class ||= Audit
     end
 
+    def enable_auditing
+      store[:auditing_enabled] = nil
```
tbrisker: Shouldn't this be true?

lib/audited/auditor.rb
```diff
@@ -285,7 +285,8 @@ def require_comment
       def auditing_enabled
         return run_conditional_check(audited_options[:if]) &&
           run_conditional_check(audited_options[:unless], matching: false) &&
-          self.class.auditing_enabled
+          self.class.auditing_enabled &&
+          Audited.auditing_enabled?
```
tbrisker: this is redundant, we already check this in the class

fatkodima:
Updated with your suggestions. Yes, you're right. Just `attr_accessor` should be enough.

PR #428, opened 2018-03-30
Add `all_audits` method to auditable models

### LINKED ISSUE #114: Join audits and associated_audits with a single query
Is it possible to join an object's audits and its associated audits into a single database query?  Right now it seems the only way to do it is by building both as arrays and combining them. It would be nice to do that in the database.  

The method now:

``` ruby
class MyObject
  audited
  has_associated_audits

  def all_audits
    (audits + associated_audits).sort_by(&:created_at)
  end
end
```

That just operates on the resulting array in memory. 

I'd rather do something like this:

``` ruby
o = MyObject.first
o.all_audits.reorder(:created_at).where(:some_other_condition)
```

Is there a way to do that now? Could it be something added to this gem?

### LINKED ISSUE #117: Scope for both regular and related audits
When displaying the history of an object we've found it useful to show a combination of both an object's audits and any related audits in one stream.  So far, we've had to add a scope to each class where we want this functionality, but it seems like something that would be great if the gem provided it.  Here's an example scope:

``` ruby
  def all_audits
    Audited.audit_class.unscoped.
        where('(associated_type = :type AND associated_id = :id) OR (auditable_type = :type AND auditable_id = :id)', type: 'User', id: id).
        order('created_at DESC')
  end
```

fatkodima:
Closes #114 
Closes #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Added
 
 - Add ability to globally disable auditing
   [#426](https://github.com/collectiveidea/audited/pull/426)
+- Add `all_audits` method to auditable models
+  [#428](https://github.com/collectiveidea/audited/pull/428)
 
 Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Breaking changes
 
-- removed `audited_columns`, `non_audited_columns`, `auditing_enabled=` instance methods, 
+- removed `audited_columns`, `non_audited_columns`, `auditing_enabled=` instance methods,
   use class methods instead
   [#424](https://github.com/collectiveidea/audited/pull/424)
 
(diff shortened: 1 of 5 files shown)


fatkodima:
Updated with new name. `own_and_associated_audits` looks pretty clear 👍

Kani999:
@tbrisker When it will be released?

tbrisker:
@Kani999 it has been released in 4.8.0 in August.

Kani999:
@tbrisker - I've been confused because of the changelog.

https://github.com/collectiveidea/audited/blob/master/CHANGELOG.md#unreleased

tbrisker:
@Kani999 ah, you're right, looks like the changelog in the master branch didn't get updated with the 4.8.0 changes. PRs welcome :wink:

PR #429, opened 2018-03-31
Ensure enum changes are stored consistently

### LINKED ISSUE #202: Enum field change
Hello,

I have status field, that has enum type in ActiveRecord. When value of this field changes, old_value is string, but new_value is integer.

Is it bug or feature?

### LINKED ISSUE #214: [More to a Upgrade] "enum" fields
I use "enum" fields (rails don't really have enums but I got a enum class with some constants). And the audit saves the numeric value that was saved in the database, but it would be better to save the description of that field.

Would be a better way to save the data as a hash. So I can get the hash and treat it before showing back to the users

### LINKED ISSUE #412: Strange new entry in audited_changes column in 4.6.0
I noticed this in an audit record that was just created in my app. Just a normal update of an audited model. 4.5.0 didn't do it. Is this indication of a problem? I do know that Rails 5.1 deprecated use of HashWithIndifferentAccess in favor of ActiveSupport:HashWithIndifferentAccess.

image

fatkodima:
```ruby
enum status: { new_status: "New", active: "In progress", complete: "Complete" }
```
Currently, when record is saved or destroyed, their audits contain enum values ("New", "In progress" or "Complete"). So I fixed "update" audits to store enum values as well for consistency. Also storing enum values (but not keys) allows to easily change enum key namings in future without changing existing audits. 

Fixes #202 
Fixes #214 
Fixes #412

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -142,7 +142,8 @@ module Audited
 
       # List of attributes that are audited.
       def audited_attributes
-        attributes.except(*self.class.non_audited_columns)
+        audited_attributes = attributes.except(*self.class.non_audited_columns)
+        normalize_enum_changes(audited_attributes)
       end
 
       # Returns a list combined of record audits and associated audits.
@@ -196,11 +197,35 @@ module Audited
 
       def audited_changes
         all_changes = respond_to?(:changes_to_save) ? changes_to_save : changes
-        if audited_options[:only].present?
-          all_changes.slice(*self.class.audited_columns)
-        else
-          all_changes.except(*self.class.non_audited_columns)
+        filtered_changes = \
+          if audited_options[:only].present?
+            all_changes.slice(*self.class.audited_columns)
+          else
+            all_changes.except(*self.class.non_audited_columns)
+          end
+
+        filtered_changes = normalize_enum_changes(filtered_changes)
+        filtered_changes.to_hash
+      end
+
+      def normalize_enum_changes(changes)
+        self.class.defined_enums.each do |name, values|
+          if changes.has_key?(name)
+            changes[name] = \
+              if changes[name].is_a?(Array)
+                changes[name].map { |v| values[v] }
+              elsif rails_below?('5.0')
+                changes[name]
+              else
+                values[changes[name]]
+              end
+          end
         end
+        changes
+      end
+
+      def rails_below?(rails_version)
+        Gem::Version.new(Rails::VERSION::STRING) < Gem::Version.new(rails_version)
       end
 
       def audits_to(version = nil)
--- a/spec/audited/auditor_spec.rb
+++ b/spec/audited/auditor_spec.rb
@@ -248,7 +248,7 @@ describe Audited::Auditor do
   end
 
   describe "on create" do
-    let( :user ) { create_user audit_comment: "Create" }
+    let( :user ) { create_user status: :reliable, audit_comment: "Create" }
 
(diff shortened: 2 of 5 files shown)


lib/audited/auditor.rb
```diff
@@ -196,10 +196,22 @@ def revision_with(attributes)
...
+          else
+            all_changes.except(*self.class.non_audited_columns)
+          end
+
+        filtered_changes = normalize_enum_changes(filtered_changes)
+        filtered_changes.to_hash
+      end
+
+      def normalize_enum_changes(changes)
+        changes.each do |key, value|
```
tbrisker: instead of looping over all the changes, how about looping on the definded enums? in most cases there would be 0-2 enums while changes could be many more.

lib/audited/auditor.rb
```diff
@@ -196,10 +196,22 @@ def revision_with(attributes)
...
-          all_changes.except(*self.class.non_audited_columns)
+        filtered_changes = \
+          if audited_options[:only].present?
+            all_changes.slice(*self.class.audited_columns)
+          else
+            all_changes.except(*self.class.non_audited_columns)
+          end
+
+        filtered_changes = normalize_enum_changes(filtered_changes)
+        filtered_changes.to_hash
```
tbrisker: why the `to_hash`? shouldn't this already be a hash?

fatkodima: This is a hash-like object (`ActiveSupport::HashWithIndifferentAccess`)

fatkodima:
Updated.

spec/audited/auditor_spec.rb
```diff
@@ -290,7 +290,7 @@ def non_column_attr=(val)
 
   describe "on update" do
     before do
-      @user = create_user( name: 'Brandon', audit_comment: 'Update' )
+      @user = create_user( name: 'Brandon', status: 0, audit_comment: 'Update' )
```
tbrisker: Shouldn't this be `status: :active` to ensure that is converted correctly?

fatkodima:
@tbrisker done

lib/audited/auditor.rb
```diff
@@ -196,11 +197,35 @@ def revision_with(attributes)
...
+      end
+
+      def normalize_enum_changes(changes)
+        self.class.defined_enums.each do |name, values|
+          if changes.has_key?(name)
+            changes[name] = \
+              if changes[name].is_a?(Array)
+                changes[name].map { |v| values[v] }
+              elsif rails_below?('5.0')
+                changes[name]
```
tbrisker: Why for rails<5 this is a noop but only if the change isn't an array?

fatkodima: That is due to the fact, that rails stores enum changes differently for rails < 5 and for rails >= 5.

travisofthenorth:
@tbrisker any idea when this will be in a published version?

lib/audited/auditor.rb
DiannaTz: @fatkodima a little late to the party, but this object indeed was an `ActiveSupport::HashWithIndifferentAccess`, but after you call `to_hash` on it, it's not anymore, it's simply a `Hash` object.  The reason I'm writing this is that after upgrading from 4.8.0 to 4.9.0 I noticed that i cannot access some audited data by simbol, because they are not stored as `ActiveSupport::HashWithIndifferentAccess` anymore. Was this intended?

Issue #430, opened 2018-04-01
Audited 5.0 discussion - what should we drop?


tbrisker:
Since we have already merged one breaking change, perhaps this is a good opportunity to make a bunch of other breaking changes and release version 5.0?
Since we don't have any other communication channel at this time, I'm opening this issue to discuss this. (I just registered the `#audited` channel on freenode if anyone want to chat via IRC.)

I personally think we should drop support for Rails 4.x and Ruby < 2.3, but there are most likely other things we can change.
Users needing support for older versions will still be able to use the 4.x branch.

tbrisker:
@danielmorrison @fatkodima @grosser @hernanat @codenamev @lmansur @tekin @JonathanWThom - Since you have been actively contributing lately, would be happy for your inputs as well :smiley:

hernanat:
I am mobile at the moment, but I definitely do have some input for this. Will respond when I am in front of a computer, thanks for tagging me :)

grosser:
drop ruby <2.3 and start using frozen string + rubocop :)

fatkodima:
Maybe also we can reduce complexity of all logic related to db updates? For example, at least by merging all this files https://github.com/collectiveidea/audited/tree/master/lib/generators/audited/templates into one initial migration?

fatkodima:
Probably, this can be merged as well https://github.com/collectiveidea/audited/pull/406, if this is a pretty frequent usecase.

grosser:
merging the migrations sounds nice, but I'd prefer to still keep at least 2
so it's obvious what is to do when adding a new migration

grosser:
drop rails 4.0 and 4.1 support too ...
https://github.com/collectiveidea/audited/pull/431

fatkodima:
I'm also a bit concerned about usefulness of this part:
> If you're using PostgreSQL, then you can use rails generate audited:install --audited-changes-column-type jsonb (or json) to store audit changes natively with its JSON column types. If you're using something other than integer primary keys (e.g. UUID) for your User model, then you can use rails generate audited:install --audited-user-id-column-type uuid to customize the audits table user_id column type.

It is not that hard to edit generated migration file by yourself without providing those extra options.

codenamev:
Some of these would be easier than others, but here are a few features I would love to see:
- Undo multiples (maybe given a timestamp/range)
- Chained audits (thinking some kind of simulated linked-list), that could allow you to undo related side-affects from changes.
[...]

PR #431, opened 2018-04-01
drop support for deprecated rails versions

grosser:


fatkodima:
This also can be removed https://github.com/collectiveidea/audited/blob/9f6619ca48d3f8e0809ca7d02451a95b7f530be3/lib/audited/auditor.rb#L165

grosser:
done, good to go ?

fatkodima:
Looks like readme should be changed here
> Audited currently (4.x) works with Rails 5.1, 5.0 and 4.2. It may work with 4.1 and 4.0, but this is not guaranteed.

And added missing changelog entry

grosser:
helper is still needed since I'm not removing rails 4.2

grosser:
updated and green

grosser:
review plz

fatkodima:
@tbrisker Did you proposed to delete all 4.x (so 4.2 as well) rails versions, or just 4.0 and 4.1?

tbrisker:
I suggested dropping all in #430, but perhaps it would be better to split it in two and keep just 4.2 in the stable 4.x branch, considering that 4.{0,1} have not been officially supported for some time.

grosser:
don't think there is disagreement over dropping 4.0/4.1 ... just maybe 4.2
... also I'll need this as base for other PRs to not merge-conflict myself
:D

tbrisker:
Considering we haven't been supporting them officially for at least 2 years, I guess it's safe to drop. Thanks @grosser !

PR #432, opened 2018-04-02
Allow users to set types for all id columns

vassilevsky:
Hello :)

Thank you for the wonderful gem!

We use UUID for all id columns and this is something that we would need.

Let me know if anything else is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Added
 - Add ability to globally disable auditing
   [#426](https://github.com/collectiveidea/audited/pull/426)
 
+- Add ability to set all id column types with the install generator
+  [#432](https://github.com/collectiveidea/audited/pull/432)
+
 Changed
 
 - None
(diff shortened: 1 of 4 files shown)


fatkodima:
I don't think this is that hard to do it manually after migration was generated to complicate things.
So I'm 👎 on this.

vassilevsky:
Yep, it's really easy to edit the generated migration. Do you think all options can be removed then?

fatkodima:
Yes, I think so. I have proposed that here https://github.com/collectiveidea/audited/issues/430#issuecomment-377825088
but has not yet received any feedback.

josephpage:
I think it should be completely removed, but it should use the default column type from Rails.
The one which can be configured with 
```
config.generators do |g|
  g.orm :active_record, primary_key_type: :uuid
end
```
I can work on it in my PR #433

vassilevsky:
I have created #434 that completely removes options instead :)

PR #433, opened 2018-04-04
Refactor install migration for better readability

josephpage:
Notable changes :
- removed `:force => true` (now the migration will fail if a table named `audits` already exists)
- renamed indexes on polymorphic association fields : `index_audits_on_assocation` to avoid collision with already existing indexes (particularly `user_index`, too general)
- changed the default `id` column type for polymorphic association on Rails 5.1+, `bigint` instead of `integer` (that is achieved automatically by `references` method on Rails 5.1+)

Other minor changes : 
- changed migration method : from `self.up`/`self.down` to `change` (best practice)
- use of Ruby 1.9 hash syntax
- moved columns type into method name

### DIFF
--- a/lib/generators/audited/templates/install.rb
+++ b/lib/generators/audited/templates/install.rb
@@ -1,30 +1,25 @@
 class <%= migration_class_name %> < <%= migration_parent %>
-  def self.up
-    create_table :audits, :force => true do |t|
-      t.column :auditable_id, :integer
-      t.column :auditable_type, :string
-      t.column :associated_id, :integer
-      t.column :associated_type, :string
-      t.column :user_id, :<%= options[:audited_user_id_column_type] %>
-      t.column :user_type, :string
-      t.column :username, :string
-      t.column :action, :string
-      t.column :audited_changes, :<%= options[:audited_changes_column_type] %>
-      t.column :version, :integer, :default => 0
-      t.column :comment, :string
-      t.column :remote_address, :string
-      t.column :request_uuid, :string
-      t.column :created_at, :datetime
+  def change
+    create_table :audits do |t|
+      t.belongs_to :auditable, polymorphic: true, index: false
+      t.index [:auditable_type, :auditable_id, :version],
+              name: :index_audits_on_auditable_and_version
+
+      t.references :associated, polymorphic: true,
+                                index: { name: :index_audits_on_associated }
+
+      t.references :user, type: :<%= options[:audited_user_id_column_type] %>,
+                          polymorphic: true,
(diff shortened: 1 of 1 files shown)


tbrisker:
@josephpage Thank you! Looks like this has conflicts with the master branch due to a PR that was just merged, can you please rebase?

lib/generators/audited/templates/install.rb
```diff
@@ -1,30 +1,25 @@
...
-      t.column :audited_changes, :<%= options[:audited_changes_column_type] %>
-      t.column :version, :integer, :default => 0
-      t.column :comment, :string
-      t.column :remote_address, :string
-      t.column :request_uuid, :string
-      t.column :created_at, :datetime
-    end
+  def change
+    create_table :audits do |t|
+      t.belongs_to :auditable, polymorphic: true, index: false
```
fatkodima: Probably, this should be `.references` for consistency with lines below

lib/generators/audited/templates/install.rb
```diff
@@ -1,30 +1,25 @@
...
+      t.index [:auditable_type, :auditable_id, :version],
```
fatkodima: I think, all separate index definitions should be defined at the bottom (after all column definitions) of a method, as always done in rails migrations

fatkodima: And please preserve an old name

lib/generators/audited/templates/install.rb
```diff
@@ -1,30 +1,25 @@
...
+                          polymorphic: true,
+                          index: { name: :index_audits_on_user }
+
+      t.string :username
+      t.string :action
+      t.<%= options[:audited_changes_column_type] %> :audited_changes
+      t.integer :version, default: 0
+      t.string :comment
+      t.string :remote_address
+      t.string :request_uuid, index: true
```
fatkodima: If i remember correctly, all methods (like `.string`, `.datetime`) based on `.add_column` does not have `:index` option. So this should be done through `.add_index`

lib/generators/audited/templates/install.rb
josephpage: Do you prefer `auditable_index` ?

lib/generators/audited/templates/install.rb
fatkodima: https://github.com/rails/rails/issues/31281

lib/generators/audited/templates/install.rb
fatkodima: If you rename this index, this possibly will break upgrading by `rails generate audited:upgrade`

PR #435, opened 2018-04-04
bump minimum ruby version

grosser:
drop all EOL ruby versions ... enable us to use frozen strings magic comments and newer syntax

@tbrisker


audited.gemspec
```diff
@@ -5,7 +5,6 @@ require "audited/version"
 Gem::Specification.new do |gem|
   gem.name        = 'audited'
   gem.version     = Audited::VERSION
-  gem.platform    = Gem::Platform::RUBY
```
grosser: default platform, so we don't need to set it

PR #438, opened 2018-04-09
basic rubocop checks and fixes

grosser:
@tbrisker 

starter PR to later slowly add more things like frozen strings etc performance related things and maybe some styles if there is consensus

### DIFF
--- a/audited.gemspec
+++ b/audited.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |gem|
 
   gem.add_development_dependency 'appraisal'
   gem.add_development_dependency 'rails', '>= 4.2', '< 5.3'
+  gem.add_development_dependency 'rubocop', '~> 0.54.0'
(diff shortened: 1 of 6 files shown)


audited.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |gem|
 
   gem.add_development_dependency 'appraisal'
   gem.add_development_dependency 'rails', '>= 4.2', '< 5.2'
+  gem.add_development_dependency 'rubocop', '~> 0.54.0'
```
grosser: needs to be locked down to the minor since new cops gets added all the times and then would break the tests

grosser:
I want a PR that does not get a bunch of discussion/bike-shedding and has a
small changeset.

grosser:
I can haz merge ?

grosser:
@fatkodima ?

grosser:
@tbrisker rebased and ready

Issue #446, opened 2018-04-27
Audited does not correctly deserialize attributes of JSON type, specifically when the value is an array


cfeckardt:
Context:

When trying to store an array (json object) in an attribute, it returns only the last value of that array when it is audited.

Consider:
```
model Foo
  # has an attribute of type :json called my_attr
end
Foo.create(my_attr: ['hello', 'world'])

Foo.first.revision_at(0.seconds.ago).my_attr # returns 'world', expected ['hello', 'world']
```

The culprit appears to be
```
    # Returns a hash of the changed attributes with the new values
    def new_attributes
      (audited_changes || {}).inject({}.with_indifferent_access) do |attrs, (attr, values)|
        attrs[attr] = values.is_a?(Array) ? values.last : values
        attrs
      end
    end
```

In this case values is an array but we don't want to last value.

*PROPOSED SOLUTION*
Don't test for `is_a?(Array)`, but use a `class AuditArray < Array` internally instead, so that we are able to use arrays.

Thoughts?

Happy to fix this if the approach is viable to the author.

cfeckardt:
I'm not sure I follow exactly what you mean. But I think that would still not work if you create it with a json array attribute, because it'll still be read as a historical list of values, rather than just an array.

I think that the correct solution is to *always* wrap your values in an array. Even for create or destroy. Then you can just remove that check. What do you think?

tbrisker:
Wrapping all values in an array would only work for new audits, old audits will remain in the same state. The check here, for `values.is_a?(Array)` is incorrect - it should be `action == 'update'` instead.

cfeckardt:
Ah yes that makes sense, you are completely right. Ignore the PR for now. I will sort it properly tomorrow.

cfeckardt:
PR updated, ready for review @tbrisker

PR #448, opened 2018-04-30
 #446 - Added support for attributes to be arrays

cfeckardt:
This PR should solve #446 

I have tried to keep conventions similar to what you have done in the past.

The code works by checking if the action is an update when reading the new attributes.

I have not *yet* tested this extensively in my own application so please test it first. The tests are passing with green (however there are race conditions for me on the test suite on master too)

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -72,7 +72,7 @@ module Audited
     # Returns a hash of the changed attributes with the new values
     def new_attributes
       (audited_changes || {}).inject({}.with_indifferent_access) do |attrs, (attr, values)|
-        attrs[attr] = values.is_a?(Array) ? values.last : values
+        attrs[attr] = self.action == 'update' ? values.last : values
         attrs
       end
     end
@@ -80,8 +80,7 @@ module Audited
     # Returns a hash of the changed attributes with the old values
     def old_attributes
       (audited_changes || {}).inject({}.with_indifferent_access) do |attrs, (attr, values)|
-        attrs[attr] = Array(values).first
-
+        attrs[attr] = self.action == 'update' ? Array(values).first : values
         attrs
       end
     end
--- a/spec/support/active_record/schema.rb
+++ b/spec/support/active_record/schema.rb
@@ -40,6 +40,10 @@ ActiveRecord::Schema.define do
     t.column :created_at, :datetime
     t.column :updated_at, :datetime
     t.column :favourite_device, :string
+
+    if ENV['DB'] == 'POSTGRES'
+      t.column :favourite_colours, :json
+    end
   end
 
   create_table :companies do |t|
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -115,7 +115,7 @@ module Audited
       def revisions(from_version = 1)
         return [] unless audits.from_version(from_version).exists?
 
-        all_audits = audits.select([:audited_changes, :version]).to_a
+        all_audits = audits.select([:audited_changes, :version, :action]).to_a
         targeted_audits = all_audits.select { |audit| audit.version >= from_version }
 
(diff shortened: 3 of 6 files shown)


spec/support/active_record/schema.rb
```diff
@@ -40,6 +40,7 @@
     t.column :created_at, :datetime
     t.column :updated_at, :datetime
     t.column :favourite_device, :string
+    t.column :favourite_colours, :json
```
tbrisker: json is only supported on postgres, tests are failing for mysql/sqlite because of this.

cfeckardt: I ran `bundle exec rspec`. I see now that only runs tests on 1 db.

Will address this by only running the new tests if the db is postgres.

cfeckardt:
@tbrisker changed as per your requests, with a green build

spec/support/active_record/schema.rb
cfeckardt: @tbrisker The changes you requested have been implemented.

lib/audited/audit.rb
```diff
@@ -72,16 +72,15 @@ def revision
...
         attrs
       end
     end
 
     # Returns a hash of the changed attributes with the old values
     def old_attributes
       (audited_changes || {}).inject({}.with_indifferent_access) do |attrs, (attr, values)|
-        attrs[attr] = Array(values).first
-
+        attrs[attr] = self.action == 'update' ? Array(values).first : values
```
tbrisker: Not crucial, but now we don't need to wrap the values as an array since when the action is an `update` the changes are saved as an array.

spec/support/active_record/schema.rb
```diff
@@ -40,6 +40,10 @@
     t.column :created_at, :datetime
     t.column :updated_at, :datetime
     t.column :favourite_device, :string
+
+    if ENV['DB'] == 'POSTGRES'
+      t.column :favourite_colours, :json
```
tbrisker: Is there a way to test this on other DBs as well? perhaps using serialization to save the array as a string column?

lib/audited/audit.rb
cfeckardt: Sure

spec/support/active_record/schema.rb
cfeckardt: I considered this.

However, if the field is not :json, then active record does not automatically cast, and audited_changes would contains strings.

We'd get ```audited_changes: ["['blue', 'red']", "['blue']"]```
instead of ```[[blue, red],[blue]]```

Therefore, it's just another string, and already covered in other tests.

If you disagree I'd be happy to come up with something a bit more explicit

cfeckardt: @tbrisker thoughts?

tbrisker: sure, this is better in any case. I think we could use serialize somehow to test other dbs, but i won't block on that.

tbrisker:
ping @cfeckardt - care to fix the remaining comment and add a line to the changelog?

davidpatters0n:
@cfeckardt - Would love to see this finalised. Any chance on getting this over the finishing line? c @tbrisker

gingerlime:
sorry for the "me too", but I'm just trying audited out and bumped into this limitation. When I pull a revision out, it shows `nil` for `jsonb` fields that had an empty array in there. Looks like this has been open for a while. Is there anything I can help with? (I literally just started *using* audited a couple of hours ago, so not sure how much I can contribute, but happy to try)

spec/support/active_record/schema.rb
alexanderweiss: I haven't looked into it too deeply, but at least in Rails 5 this can be tested using `audit.revision`. That will raise a `SerializationTypeMismatch` when trying to get the revision for a `create` audit that contains an array serialized using Rails' `serialize` (because it gets a string rather than an array).

PR #454, opened 2018-05-17
allow private methods on audited condition

rauann:
Adding `true`as a second parameter to include private methods on conditional verification.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -285,9 +285,8 @@ module Audited
 
       def run_conditional_check(condition, matching: true)
         return true if condition.blank?
-
         return condition.call(self) == matching if condition.respond_to?(:call)
-        return send(condition) == matching if respond_to?(condition.to_sym)
+        return send(condition) == matching if respond_to?(condition.to_sym, true)
 
         true
       end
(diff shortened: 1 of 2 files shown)


lib/audited/auditor.rb
```diff
@@ -285,11 +285,10 @@ def auditing_enabled
 
       def run_conditional_check(condition, matching: true)
         return true if condition.blank?
-
         return condition.call(self) == matching if condition.respond_to?(:call)
-        return send(condition) == matching if respond_to?(condition.to_sym)
+        return send(condition) == matching if respond_to?(condition, include_all: true)
 
-        true
+        false
```
tbrisker: why did you change the default to false?

lib/audited/auditor.rb
```diff
@@ -285,11 +285,10 @@ def auditing_enabled
 
       def run_conditional_check(condition, matching: true)
         return true if condition.blank?
-
         return condition.call(self) == matching if condition.respond_to?(:call)
-        return send(condition) == matching if respond_to?(condition.to_sym)
+        return send(condition) == matching if respond_to?(condition, include_all: true)
```
tbrisker: This actually isn't a kwarg, so just `true` as the second parameter is enough. (this still works since you are creating a new hash here and passing it to the function, and since a hash is non-falsy the check in `respond_to?` still passes, but every check of this function creates an extra object allocation)

lib/audited/auditor.rb
rauann: With the second parameter as `true` on `respond_to?` at last return, the default will never be raised I think. Anyway, I'll rollback it.

lib/audited/auditor.rb
rauann: Sure, good catch. Updated.

Issue #479, opened 2018-12-17
Use bigint for foreign keys


danielmorrison:
Rails since 5.1 defaults to bigint for ID columns and we should default to the same for `auditable_id`, `associated_id` and `user_id`.

johnclittle:
Is this still being considered? I could look into it but not sure how.

PR #481, opened 2019-01-22
added bigint for ID columns

ConSou:
attempts to resolves issue #479


ConSou:
Hello, I'm not entirely sure what is causing all of the `Bundler could not find compatible versions for gem "bundler":` errors.  I see `gemfiles/rails42.gemfile` is the file that is causing all of the failures and the travis cl job log seems to point to the rails version not having the correct bundler version.  I'd love to fix this, but not sure where to start.  Any thoughts or advice?  Thanks!

danielmorrison:
Cool, i'll take a look deeper later in the week if nobody else does first. 

One thing we'll need to add is another migration for existing users.

PR #443, opened 2018-04-10
rename instance attribute version to audit_version (Part Deux)

jeffdill2:
Just taking [this pull request](https://github.com/collectiveidea/audited/pull/307) the last few inches to finish it up since it had gone stale. 😄

Issue #480, opened 2019-01-22
Kaminari's `page_method_name` method does not work on Audit model


git-venkat:
I had to override kaminari's `page_method_name` in my project as follows, 
```
Kaminari.configure do |config|
  config.page_method_name = :kaminari_page
end

```
Now calling this method on Audit model does not work,
`Audit.kaminari_page(params[:page])` throws NoMethodError (undefined method `kaminari_page' for Audit model. Where as other ActiveRecord model work fine with this method.

PR #482, opened 2019-01-30
Remove deprecation warning, remove method_missing

shaicoleman:
The method_missing might interfere with other method_missing in the class, e.g. ActiveRecord::Delegation#method_missing
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation/delegation.rb#L111

I originally considered just defining the version method, but that was in the original PR, and it was decided not to do it.
https://github.com/collectiveidea/audited/pull/443#discussion_r180718047

I think the best way forward is just to remove the deprecation warning and to do a major version bump.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -90,16 +90,6 @@ module Audited
     end
 
     module AuditedInstanceMethods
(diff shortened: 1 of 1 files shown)


shaicoleman:
Possibly related to #480

shaicoleman:
BTW: branch 4.8-stable is not merged into master

PR #486, opened 2019-02-26
Include an example of passing strings to `as_user`

stevenpetryk:
For some reason, on my first read of this README, I totally missed the fact that you could pass a string in anywhere you can pass an ActiveRecord model. Having an example like this would have saved me some silly mock-object-creation. Willing to change the copy or the specificity of the example, but I have a feeling this is a common use case.


domcleal:
Thanks @stevenpetryk!

PR #496, opened 2019-05-28
Document 2.6.3 as tested Ruby versions

jheijmans:
I was surprised there was no 2.6.x version tested, but it is actually tested, just not documented.


domcleal:
Whoops, my fault. Thanks @jheijmans!

PR #498, opened 2019-06-18
Add info about mysql 5.7+ json type to readme

Vasfed:
MySQL 5.7 supports `json` column type and rails 5 have support for that similarly to postgres.

Mention this in README along with what happens by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ $ rails generate audited:install
 $ rake db:migrate
 ```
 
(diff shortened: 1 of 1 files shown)


domcleal:
Looks good to me @Vasfed, thanks for the information! Would you mind just removing the MySQL fix please? I merged in #499 which includes the same, and don't want to duplicate the lines in the file.

You should be able to do a `git reset --hard HEAD^` locally on your branch and force-push to remove the second commit, or use `git revert`.

Vasfed:
@domcleal done, rebased the branch onto current master

domcleal:
Thanks @Vasfed! 👏

PR #502, opened 2019-07-09
Allow for Audting to be enabled temporarily

trevorrjohn:
Some models I only want auditing in certain conditions, this allows me to disable it by default and turn it on only when needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Breaking changes
 
 Added
 
+- Add `with_auditing` methods to enable temporarily
+  [#502](https://github.com/collectiveidea/audited/pull/502)
 - Add `update_with_comment_only` option to control audit creation with only comments
   [#327](https://github.com/collectiveidea/audited/pull/327)
 - Support for Rails 6.0 and Ruby 2.6
--- a/README.md
+++ b/README.md
@@ -332,6 +332,23 @@ To disable auditing on all models:
 Audited.auditing_enabled = false
 ```
 
+If you have auditing disabled by default on your model you can enable auditing
+temporarily.
+
+```ruby
+User.auditing_enabled = false
+@user.save_with_auditing
+```
+
+or:
+
+```ruby
+User.auditing_enabled = false
+@user.with_auditing do
+  @user.save
+end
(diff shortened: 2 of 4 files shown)


trevorrjohn:
I am not sure why the mysql tests are failing on travis. They are passing locally. @domcleal do you know what this might be?

README.md
```diff
@@ -329,6 +329,23 @@ To disable auditing on all models:
 Audited.auditing_enabled = false
 ```
 
+If you have audting disabled by default on your model you can enable auditing
```
domcleal: Ditto the typo:

```suggestion
If you have auditing disabled by default on your model you can enable auditing
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Breaking changes
 
 Added
 
+- Add `with_audting` methods to enable temporarily
```
domcleal: Small typo:

```suggestion
- Add `with_auditing` methods to enable temporarily
```

spec/audited/auditor_spec.rb
```diff
@@ -806,6 +806,82 @@ def stub_global_max_audits(max_audits)
...
+      end
+      t1.join
+      t2.join
+
+      Models::ActiveRecord::User.enable_auditing
+      expect(Models::ActiveRecord::User.find_by_name('Shaggy').audits.count).to eq(1)
+      expect(Models::ActiveRecord::User.find_by_name('Scooby').audits.count).to eq(0)
+    end
+
+    it "should save an audit when auditing is globally disabled" do
```
domcleal: I'm a bit unsure what this test's doing - it doesn't seem to call the new methods and while the description says "should save an audit", the checks are for `expect(user.audits.count).to eq(0)`.

Should it be using `with_auditing { }` somewhere? Perhaps around the `create_user` call and then check it creates one audit?

trevorrjohn: Yeah, this test is kind of bogus. I am going to delete it. This scenario is covered in the other specs.

domcleal:
Thanks for the PR @trevorrjohn 💯

Issue #515, opened 2019-09-09
redacted is not applied on create


francescob:
when saving the audit for a create action, fields marked as redacted have their value saved in the database, on update the value is correctly ignored

JonathanWThom:
I was able to replicate this. Unrelated, but I also noticed that it includes the redacted element (redacted though) in the audited_changes column whether it was changed or not, which it probably shouldn't. I was responsible for the redacted feature, so will do my best to fix this.

vlad-psh:
Fixed in [audited 5.1.0](https://github.com/collectiveidea/audited/blob/main/CHANGELOG.md#510-2022-12-23) (2022-12-23) via #630

PR #518, opened 2019-09-17
Redacts values on create

JonathanWThom:
* https://github.com/collectiveidea/audited/commit/c200bb5647872609567215f2eaaa0ff1e2cb1662 added the `redacted` feature, which should redact chosen attributes while not ignoring them completely. It was previously only working on update; this commit makes it work on create as well.

* Additionally, redacted attributes (with the correct redaction value) were showing up in `audited_changes` column, even if they weren't changed. This fixes that as well.

Resolves https://github.com/collectiveidea/audited/issues/515

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -101,7 +101,7 @@ module Audited
 
     module AuditedInstanceMethods
       REDACTED = '[REDACTED]'
-      
+
       # Temporarily turns off auditing while saving.
       def save_without_auditing
         without_auditing { save }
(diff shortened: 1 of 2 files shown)


joshforbes:
Thanks for this. 🙇 I need this in order to start using this gem.

JonathanWThom:
Welcome! @joshforbes

dzunk:
Odds of getting this fix merged soon, and maybe a patch release? The redaction feature released in 4.10.0 is pretty broken without this PR.

commitshappen:
+1

JonathanWThom:
Sounds like this was resolved in https://github.com/collectiveidea/audited/pull/630

PR #521, opened 2019-09-28
Support multiple databases

michael-groble:
We would like to store the `audits` table in a separate database.  With Rails 6 or [ankane/multiverse](https://github.com/ankane/multiverse), I can do that by setting up an abstract base record for all models I want to store in a secondary database:

```ruby
class SecondaryRecord < ActiveRecord::Base
  self.abstract_class = true
  connects_to database: { writing: :secondary, reading: :secondary_replica }
end
```
and would then want to set up the audit model to live there:
```ruby
class CustomAudit < SecondaryRecord
end
```

But since that `CustomAudit` record does not inherit from `Audited::Audit`, this won't really work.

This pull request creates an `Audited::Auditable` concern which lets me store audits in the secondary database like so:

```ruby
class CustomAudit < SecondaryRecord
  include Audited::Auditable
end
```

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -33,28 +33,83 @@ module Audited
     end
   end
 
-  class Audit < ::ActiveRecord::Base
-    belongs_to :auditable,  polymorphic: true
-    belongs_to :user,       polymorphic: true
-    belongs_to :associated, polymorphic: true
+  module Auditable
+    extend ::ActiveSupport::Concern
 
-    before_create :set_version_number, :set_audit_user, :set_request_uuid, :set_remote_address
+    included do
+      belongs_to :auditable,  polymorphic: true
+      belongs_to :user,       polymorphic: true
+      belongs_to :associated, polymorphic: true
 
-    cattr_accessor :audited_class_names
-    self.audited_class_names = Set.new
+      before_create :set_version_number, :set_audit_user, :set_request_uuid, :set_remote_address
 
-    serialize :audited_changes, YAMLIfTextColumnType
+      cattr_accessor :audited_class_names
+      self.audited_class_names = Set.new
 
-    scope :ascending,     ->{ reorder(version: :asc) }
-    scope :descending,    ->{ reorder(version: :desc)}
-    scope :creates,       ->{ where(action: 'create')}
-    scope :updates,       ->{ where(action: 'update')}
-    scope :destroys,      ->{ where(action: 'destroy')}
+      serialize :audited_changes, YAMLIfTextColumnType
 
-    scope :up_until,      ->(date_or_time){ where("created_at <= ?", date_or_time) }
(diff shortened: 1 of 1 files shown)


brand-it:
```
NotImplementedError:
  Audit is an abstract class and cannot be instantiated.
  # /Users/blareau/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/activerecord-6.0.2/lib/active_record/inheritance.rb:52:in `new'
```
@michael-groble Did you get the error above and if so how did you work around the issue?

michael-groble:
I see failures in travis, @brand-it, but things run cleanly for me locally (on macos), e.g.:
```bash
bundle exec rake
bundle exec appraisal rails50 rake
bundle exec appraisal rails51 rake
bundle exec appraisal rails52 rake
bundle exec appraisal rails60 rake
```
(and the travis errors don't seem to match that abstract class error)

LukeGuanY:
Hi @michael-groble, I also have the same need that writing the audit logs to the second database. But I am having trouble with how to tell the CustomAudit to write the audit logs to the second database. Could you please tell me how to point audited to the second DB in the CustomAudit file?

michael-groble:
@LukeGuanY , that really depends on how you connect to multiple databases in the first place.  I give an example above of rails 6.  You can also set up a base that connects to a specific entry in you database configuration.  For example, say your database config has a separate set of configuration for "warehouse" type tables, e.g.

```
production:
  ...
  database: app_production
  warehouse:
    ...
    database: warehouse_production
```
then you can have a base record (rails 5 this time)
```
class WarehouseRecord < ApplicationRecord
  self.abstract_class = true
  establish_connection configurations[Rails.env]['warehouse'] if configurations[Rails.env]['warehouse']
end

class CustomAudit < WarehouseRecord
  include Audited::Auditable
end
```

LukeGuanY:
@michael-groble Thanks for the reply. I really appreciate your method. I am sharing my configuration here so that people can also learn it. 

I use the multiple database feature on Rail 6

Database.yml:
```
development:
  primary:
    database: the_existing_db
    ...
  audit:
    database: audit_db
    migrations_paths: db/audit_migrate #you need to create a new folder called "audit_migrate" under db directory
```

Make an audit initializer at  `config/initializers/audited.rb` 
```
# Make Audited send its data to a different database
Audited::Audit.class_eval do
  establish_connection :audit
end
```

vinayuttam:
The above suggestion works fine up to version `v5.0.1`.  With the recent update `v5.0.2`, we get the following error
```
config/initializers/audited.rb:3:in `<main>': uninitialized constant Audited::Audit (NameError)
Did you mean?  Audited::Auditor
```

amkisko:
Probably related to this one https://github.com/collectiveidea/audited/pull/641

Issue #517, opened 2019-09-16
New enum handling in 4.9.0 is a breaking change


gaffney:
The enum change (#429) is a major breaking change but is not listed under `Breaking Changes` (data in the database still contains the enum string values).

Ideally the `CHANGELOG` would list the enum change as such, and provide a migration path for users who already rely on the behavior of `4.8.0`.

Also, semi-related - The [CHANGELOG](https://github.com/collectiveidea/audited/blob/master/CHANGELOG.md) still says `4.9.0` is unreleased even though it's in [releases](https://github.com/collectiveidea/audited/releases) (or if it's trying to show nothing under UNRELEASED, that is not very clear).

cc @fatkodima 

-----

**EDIT**: Deleted some paragraph saying that reverting #429 still produced different results - that was due to a bad merge conflict resolution 🤦‍♂ Reverting the PR fixes the issue.

travisofthenorth:
Can someone from collectiveidea chime in here? @danielmorrison 

This is preventing upgrades to Rails 6 in a number of my applications. I have a candidate PR ready to go to allow the previous behavior: https://github.com/collectiveidea/audited/pull/526

I also have another PR to fix the build, which is currently failing: https://github.com/collectiveidea/audited/pull/530

danielmorrison:
I'll plan to take a look later this week.

gaffney:
Bump, any update @danielmorrison ?

kluka-CS:
Thanks a lot for this report and the quick reversion of the feature! Prevented huge confusion for anyone reading our database.

PR #526, opened 2020-01-27
Restore previous enum behavior with flag

travisofthenorth:
Resolves https://github.com/collectiveidea/audited/issues/517 with a configuration flag to restore the previous behavior

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -2,7 +2,12 @@ require 'active_record'
 
 module Audited
   class << self
-    attr_accessor :ignored_attributes, :current_user_method, :max_audits, :auditing_enabled
+    attr_accessor \
+      :auditing_enabled,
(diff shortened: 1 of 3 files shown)


audited.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'rubocop', '~> 0.54.0'
   gem.add_development_dependency 'rspec-rails', '~> 3.5'
   gem.add_development_dependency 'single_cov'
+  gem.add_development_dependency 'sprockets', '< 4.0'
```
travisofthenorth: This fixes the build. I'm not familiar with how the test rails app works, so I just pinned this for now.

danielmorrison (approved):

PR #530, opened 2020-04-19
Fix build issues

travisofthenorth:
- Install a 1.x version of Bundler for Rails < 5
- Insert an empty sprockets manifest file for sprockets >= 4
- Migration file fix for Rails 6

PR #532, opened 2020-05-12
Follow Rails autoloading conventions

duncanjbrown:
Depends on #530, because it won't pass specs without.

Thank you for this excellent gem, which we're using happily in production.

I encountered a conflict between `audited` and another gem which I believe is caused by the way audited hooks into Rails core classes. A minimal app displaying the behaviour is here: https://github.com/duncanjbrown/audited-bug. The behaviour is outlined on a rubyonrails-talk thread here: https://discuss.rubyonrails.org/t/understanding-a-conflict-between-two-activesupport-on-load-calls/74836.

Where an `on_load` reference is required to a non-ActiveSupport class (such as `ActionController::API`), wrapping the initialization code in a proper `railtie` prevents the weirdness I see in the above app. As far as I can tell this is the correct way to inject this code and it doesn't seem to break `audited`, which is why I'm submitting this PR.

While investigating this I learned that the current guidance (https://edgeguides.rubyonrails.org/engines.html#avoid-loading-rails-frameworks) it is best practice to extend core classes by hooking into ActiveSupport.on_load hooks, so this PR includes a change to the way `audited` does that, too.

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -30,3 +30,4 @@ require 'audited/audit'
 ::ActiveRecord::Base.send :include, Audited::Auditor
 
 require 'audited/sweeper'
+require 'audited/railtie'
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -27,7 +27,9 @@ end
 require 'audited/auditor'
 require 'audited/audit'
 
-::ActiveRecord::Base.send :include, Audited::Auditor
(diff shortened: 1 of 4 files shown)


luke-hawk:
This will also fix an issue I encountered when using audited with [devise-jwt](https://github.com/waiting-for-dev/devise-jwt). Thanks for investigating and fixing @duncanjbrown !

PR #527, opened 2020-01-30
Drop Rails 4.2 support

travisofthenorth:
Title says it all

PR #533, opened 2020-05-15
own_and_associated_audits for STI

eric-hemasystems:
When `own_and_associated_audits` was introduced in PR #428 it did not take STI models into account. This PR fixes the problem.

I got an error when trying to run the test suite locally. This error also affects many of the CI builds. 72da59f fixes that issue and although unrelated to my goal is actually most of the changes for this PR. See the commit message for more info.

f5d9899 adds some tests that duplicate the problem and 489224e is the very small change to fix the issue.


eric-hemasystems:
Looking through the CI results I see:

* All the Rails 4.2 failed due to the appraisal gem not being compatible. I suggest PR #527 rather than trying to fix it as 4.2 is outside of [Rails' own maintenance policy](https://guides.rubyonrails.org/maintenance_policy.html).
* Rails 6 on PG is failing due to [the MigrationContext initializer](https://github.com/rails/rails/commit/7cc27d749c3563e6b278ad01d233cb92ea3b7935#diff-8d3c44120f7b67ff79e2fbe6a40d0ad6R1018) getting an additional argument for Rails 6. This is probably really considered an internal API for Rails so the `audited` gem should probably just adapt but this seems outside of the scope of this PR.

eric-hemasystems:
Any feedback on this. Just updated to the latest on master. Would love to get this merged so I don't have to maintain my own branch. 😉

eric-hemasystems:
Looking at CI results I don't see anything related to this PR. Seems CI related.

PR #543, opened 2020-08-11
added check to audit limit

gvn182:


lib/audited/audit.rb
```diff
@@ -171,7 +171,10 @@ def set_version_number
       if action == 'create'
         self.version = 1
       else
-        max = self.class.auditable_finder(auditable_id, auditable_type).maximum(:version) || 0
+        max = 0
+        if ::Audited.max_audits.present? && ::Audited.max_audits > 0  
+          max = self.class.auditable_finder(auditable_id, auditable_type).maximum(:version) || 0
+        end
         self.version = max + 1
```
JerrodCarpenter, line 178: I don't know if this quite works as it'd always set the version number to 1.

PR #545, opened 2020-08-28
Remove rc from Rails 6 gemfile

okuramasafumi:
This makes it a little bit clearer that it supports Rails 6.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -31,7 +31,7 @@ appraise 'rails52' do
 end
 
 appraise 'rails60' do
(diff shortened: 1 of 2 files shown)


sinsoku:
@okuramasafumi You should also modify the "Appraisals" file as well.
https://github.com/collectiveidea/audited/blob/4ad3a9c2a7dfacedc52f9f0c2a47f7b57a73e26e/Appraisals#L33-L38

okuramasafumi:
Oops I noticed that this PR is redundant now.

PR #554, opened 2020-12-10
Support for Rails 6.1

yknx4:
Required changes to support Rails 6.1

- Remove deprecated method calls
- Update Gemspec
- Update Travis Config

PR #556, opened 2020-12-11
Add support for Rails 6.1

ramblex:
Hi, I've had a go at putting in support for Rails 6.1, adding a new rails61 Appraisal and gemfile.

- Fixed deprecation warning when getting ActiveRecord::Base configuration in spec.
- Refactored YAMLIfTextColumnType to allow audited_changes type to be mocked in specs. columns_hash is now frozen and therefore, can't be mocked directly.


ramblex:
Looks like someone else is already on this! :+1:  https://github.com/collectiveidea/audited/pull/554

PR #485, opened 2019-02-16
Adds redacted option

JonathanWThom:
* In order to record that changes occurred, without storing sensitive
values, pass the `redacted` option.

* Redacted values default to `'[REDACTED]'` but can be customized with
the `redaction_value` option.

```
class User
    audited redacted: [:password, :ssn], redaction_value: SecureRandom.uuid
end
```

* A lot of this was based on the work done [here](https://github.com/collectiveidea/audited/pull/339).

* Resolves https://github.com/collectiveidea/audited/issues/475.

PR #562, opened 2021-01-12
Don't redact values when there are no changes to save

dzunk:
The redaction functionality introduced in https://github.com/collectiveidea/audited/pull/485 and released in version 4.10.0 is recording a redaction on every save, regardless of whether or not the redacted attribute has been changed.

Before:
```ruby
[4] pry(main)> model.changes_to_save
=> {}
[5] pry(main)> model.send(:audited_changes)
=> {"external_credentials"=>"[REDACTED]"}
```
After:
```ruby
[2] pry(main)> model.changes_to_save
=> {}
[3] pry(main)> model.send(:audited_changes)
=> {}
[4] pry(main)> model.external_credentials = { foo: 'bar' }
=> {:foo=>"bar"}
[5] pry(main)> model.send(:audited_changes)
=> {"external_credentials"=>["[REDACTED]", "[REDACTED]"]}
```

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -254,6 +254,7 @@ module Audited
 
       def redact_values(filtered_changes)
         [audited_options[:redacted]].flatten.compact.each do |option|
+          next unless filtered_changes.key?(option.to_s)
           changes = filtered_changes[option.to_s]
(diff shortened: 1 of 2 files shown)


dzunk:
Whoops, might be a dupe of https://github.com/collectiveidea/audited/pull/562

PR #568, opened 2021-02-22
Use Thread local variables instead of Fibers

arathunku:
GraphQL added a [fiber-based batch loading API](https://github.com/rmosolgo/graphql-ruby/pull/3264) in one of their recent releases. 

Currently, Audited uses Fiber local state instead of Thread local state and is missing information during an audit.

I've fixed this by adding specs and changing store to Thread.

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -10,7 +10,13 @@ module Audited
     end
 
     def store
-      Thread.current[:audited_store] ||= {}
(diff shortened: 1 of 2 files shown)


arathunku:
There are CI failures but they don't look related?

arathunku:
thanks @danielmorrison!

PR #576, opened 2021-06-10
Fix array support

yuki24:
This is a revised version of https://github.com/collectiveidea/audited/pull/448, which addresses https://github.com/collectiveidea/audited/issues/446 and https://github.com/collectiveidea/audited/issues/191.

The main differences are:

1. Based off of the latest commit on `master`
2. Uses `serialize :phone_numbers, Array` to replicate the issue

The rest is pretty much the same. Please let me know if there's anything you need.

cc @cfeckardt


yuki24:
This probably needs to include changes for the `#old_attributes` method as well. I'll add it shortly.

yuki24:
Thanks so much for quickly merging it!

PR #577, opened 2021-06-10
Remove the unused old_attributes method

yuki24:
While I was looking into https://github.com/collectiveidea/audited/pull/576 I noticed that the `old_attributes` method is actually unused, so I simply deleted it. I might be missing something, so feel free to close it if this was actually left in the code intentionally.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -80,13 +80,6 @@ module Audited
       end
     end
 
-    # Returns a hash of the changed attributes with the old values
-    def old_attributes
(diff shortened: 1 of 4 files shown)


yuki24:
Actually this is probably a public API on the `Audited::Audit` model rather than being an internal API, so we probably can't remove it. Closing. I'm sorry for the noise.

danielmorrison:
> Actually this is probably a public API on the `Audited::Audit` model rather than being an internal API, so we probably can't remove it. Closing. I'm sorry for the noise.

All good.

PR #580, opened 2021-06-10
Update travis.yml to reflect the current state of the Ruby/Rails versions

yuki24:
I'm basically re-sending #579. I know this includes all the commits from #574 but I just wanted to be a bit more helpful and give a ✅ so you can more easily merge the PR and keep the master branch green.

Let me know if there is anything I could do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: ruby
 cache: bundler
 rvm:
-  - 2.3.7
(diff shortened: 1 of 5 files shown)


yuki24:
The standardrb fixes need to be in every PR, so it's probably a good idea to address it first: https://github.com/collectiveidea/audited/pull/578

PR #592, opened 2021-07-27
Resolve lingering issue from collectiveidea/audited#532

mvastola:
While PR collectiveidea/audited#532 worked to prevent `audited` from initializing certain rails components too early, an issue remains wherein the gem entrypoint, `lib/audited.rb`, immediately requires `lib/audited/audit.rb`.

Therein, `Audited::Audit` subclasses, `ActiveRecord::Base` and thereby causes it to load too early in the Rails boot process.

This simple change merely moves the `require` of `audited/audit` into the `AR` `on_load` block.

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -39,9 +39,9 @@ module Audited
 end
 
 require "audited/auditor"
-require "audited/audit"
 
 ActiveSupport.on_load :active_record do
+  require "audited/audit"
   include Audited::Auditor
 end
 


akostadinov:
hmm, I'm gonna check if this fixes #599

mvastola:
> hmm, I'm gonna check if this fixes #599

Yup. It will. I didn't want to write an essay for the PR description, but this load order issue has been a huuuge pain for us, and it sucks there's no system in place to warn when configs start getting discarded.

dmke:
I've stumbled upon this, because after upgrading from 5.0.1 to 5.0.2 the specs of a Rails engine threw `uninitialized constant: Audited::Audit` errors.

The engine basically does this:

```rb
# lib/foo/engine.rb
module Foo
  class Engine < ::Rails::Engine
    require "audited"

    initializer "set audited audit class" do
      Audited.config do |config|
        config.audit_class = Foo::Audit
      end
    end
  end
end

# app/models/foo/audit.rb
class Foo::Audit < Audited::Audit
  # do some multi-tenancy shenanigans
end
```

My current workaround is to just add an `require "audited/audit"` into the engine:

```diff
   class Engine < ::Rails::Engine
     require "audited"
+    require "audited/audit"
```

The specs now run again, and the apps using the engine seem to be stable. But I do wonder, is this the right approach?

akostadinov:
@dmke , I just keep audited automatically loaded by bundler. i.e. don't add `require: false` in `Gemfile`.

dmke:
> I just keep audited automatically loaded by bundler. i.e. don't add `require: false` in `Gemfile`.

There's no `require: false`, Audited is part of the engine's `.gemspec`.

> Do you have any initializers that refer to `Audited::Audit`? You may need to enclose them in a `ActiveSupport.on_load :active_record do` block.

Yes, the engine initializer code runs in that phase, IIRC. Let me try your suggestion:

```rb
initializer "set audited audit class" do
  ActiveSupport.on_load :active_record do
    Audited.config do |config|
      config.audit_class = Foo::Audit
    end
  end
end
```

PR #607, opened 2021-12-05
Support method symbols in max_audits

gathuku:
Allow passing a method symbol to `max_audits` for dynamic maximum audits 

Example

```rb
max_audits: :max 

private 

def max
 10
end
```

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -336,6 +336,7 @@ module Audited
 
       def combine_audits_if_needed
         max_audits = audited_options[:max_audits]
+        max_audits = Integer(method(max_audits).call).abs if max_audits.is_a?(Symbol)
         if max_audits && (extra_count = audits.count - max_audits) > 0
           audits_to_combine = audits.limit(extra_count + 1)
           combine_audits(audits_to_combine)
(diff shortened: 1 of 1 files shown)


lib/audited/auditor.rb
```diff
@@ -336,6 +336,7 @@ def comment_required_state?
 
       def combine_audits_if_needed
         max_audits = audited_options[:max_audits]
+        max_audits = Integer(method(max_audits).call).abs if max_audits.is_a?(Symbol)
```
fractaledmind, line 339: Instead of `is_a?(Symbol)` check more directly `respond_to?(max_audits)`. This checks if the value of `max_audits` is a method in the calling context

lib/audited/auditor.rb
```diff
@@ -454,6 +455,8 @@ def normalize_audited_options
         audited_options[:only] = Array.wrap(audited_options[:only]).map(&:to_s)
         audited_options[:except] = Array.wrap(audited_options[:except]).map(&:to_s)
         max_audits = audited_options[:max_audits] || Audited.max_audits
+        return if max_audits.is_a?(Symbol)
```
fractaledmind, line 458: Same here. Use the more direct method check

gathuku:
@fractaledmind Thank you for the good review

PR #609, opened 2021-12-24
Use string class name and constantize

### LINKED ISSUE #608: Uninitialized Constant when defining an audit_class in the initializer
When defining a custom class in Rails 7, I get an `Uninitialized constant` error.

I believe this is because at the point of initialization, Zeitwerk hasn't been able to load the classes in.

Possible solution (others, such as Devise have done this) might be to refactor the config option to a string, and then `constantize` it when it's required in the code?

simmerz:
Fixes #608

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -13,7 +13,11 @@ module Audited
     attr_writer :audit_class
 
     def audit_class
-      @audit_class ||= Audit
+      @audit_class ||= "Audited::Audit"
+    end
+
+    def audit_model
+      audit_class.constantize
     end
 
     def store
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Audited.current_user_method = :authenticated_user
 Outside of a request, Audited can still record the user with the `as_user` method:
 
 ```ruby
-Audited.audit_class.as_user(User.find(1)) do
+Audited.audit_model.as_user(User.find(1)) do
   post.update!(title: "Hello, world!")
 end
 post.audits.last.user # => #<User id: 1>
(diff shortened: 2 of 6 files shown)


simmerz:
I'm going to reverse the quote amends. Sorry! Automatic rubocop settings!

simmerz:
I've made that change. In order to make it work, I've added a method `audit_model` which constantizes the string config value. Tests are updated to suit. Hopefully this meets with your approval

jasper502:
I second this PR - I am hitting this issue also.

jasper502:
I forked this and tried and tried it in my app - same error persists `NameError: uninitialized constant AuditTrail`.  I created a custom model:

```
class AuditTrail < Audited::Audit
  acts_as_tenant :account
end
```
then set my config:

```
Audited.config do |config|
  config.audit_class = AuditTrail
end
```

kamil98:
> I forked this and tried and tried it in my app - same error persists `NameError: uninitialized constant AuditTrail`. I created a custom model:
> 
> ```
> class AuditTrail < Audited::Audit
>   acts_as_tenant :account
> end
> ```
> 
> then set my config:
> 
> ```
> Audited.config do |config|
>   config.audit_class = AuditTrail
> end
> ```

@jasper502 it should be a string, not a class: `  config.audit_class = "AuditTrail"`

Sorc96:
I just encountered this problem as well. Thank you for the fix, hoping it gets merged.

nywton (approved):

phortx:
What's the state of this MR? The workaround doesn't seem to work. Any plans on merging this?

TvL2386:
I'm also being hit by this unfortunately....

My workaround, which is nasty, is that I just monkey patched the Audited::Audit class with my CustomAudit methods. 
So I don't have a CustomAudit anymore. At least this won't block me for now with going forward to rails-7.

I redefined Audited::Audit in my config/initializers/audited.rb

simmerz:
@nywton what needs to happen to get this merged?

nattfodd:
@danielmorrison could you please check if this can be merged along with a new version released?

lib/audited.rb
```diff
@@ -13,7 +13,11 @@ class << self
     attr_writer :audit_class
 
     def audit_class
-      @audit_class ||= Audit
+      @audit_class ||= "Audited::Audit"
```
ekohl, line 16: I think this is actually a breaking change and should not have been released as a minor version. It's really odd to have a method named `class` return a string.

PR #622, opened 2022-04-04
Make several columns in the audits table NOT NULL

### LINKED ISSUE #572: Add null constraints to migration file?
Is there a reason to why the `auditable_id`, `auditable_type`, `action`, `audited_changes`, `version`, `request_uuid`, and `created_at` fields are nullable? Are there cases where they are indeed created with null values or is it simply to be more flexible in upcoming changes without requiring the user to run a new schema migration?

jdufresne:
Improve data integrity by adding NOT NULL constraints to columns that
should _always_ have a value. This helps catch application and library
bugs earlier by catching mishandling of data.

Fixes #572


danielmorrison:
I like it. 

A couple things: 

We test migrations in `test/upgrade_generator_test.rb` and should do that here too. 

I'm confused why you made the test changes. I see that it fails without them but I'm not sure why on first glance. Seems like a smell, but maybe just a test fluke.

jdufresne:
> I'm confused why you made the test changes. I see that it fails without them but I'm not sure why on first glance. Seems like a smell, but maybe just a test fluke.

Good thinking. I tracked this down to `auditable_id` being NOT NULL, so I've reverted that portion of the change for now.

jdufresne:
@danielmorrison Any thoughts on moving ahead on this?

PR #628, opened 2022-06-23
Fix redacted attributes bug when no changes were made

vlad-psh:
I noticed two bugs with `redacted` config option. This PR fixes them both.

## Bug 1: Audited saves changes of attributes specified with `redacted:` option even if there was no changes

Given following simple configuration:

```ruby
class User
  audited redacted: :password
end
```

```
User.last.update(name: 'John')
User.last.audits.audited_changes
=> {"name"=>["Bob", "John"], "password"=>"[REDACTED]"}
```

In this example, `password` wasn't changed, but it appears in the `audited_changes`.
What is worse, `password` contains single value, while `update` actions are expected to have an array `[prev_value, new_value]`

## Bug 2: `redacted:` configuration is ignored on create and destroy actions

```
User.create(name: 'John', password: '123')
User.last.audits.audited_changes
=> {"name"=>"John","password"=>"123"}
```

In this example, password is not replaced to `[REDACTED]` placeholder.
Same for `delete` action.

PS: I also refactored `redact_values` method a little bit, so it will be a little bit easier to read, but the most important is this single line: `next unless filtered_changes.has_key?(redacted_attr)`

UPD: I found few PRs solving same issue (some of them have unsolved conflicts with master, some of them were just closed as a dupe, but none got any response since 2019) Would be great if we can fix this bug already...

### DIFF
--- a/spec/audited/auditor_spec.rb
+++ b/spec/audited/auditor_spec.rb
@@ -218,17 +218,25 @@ describe Audited::Auditor do
       redacted = Audited::Auditor::AuditedInstanceMethods::REDACTED
       user =
         Models::ActiveRecord::UserMultipleRedactedAttributes.create(
-          password: "password",
-          ssn: 123456789
+          password: "password"
         )
       user.save!
       expect(user.audits.last.audited_changes["password"]).to eq(redacted)
+      # Saving '[REDACTED]' value for 'ssn' even if value wasn't set explicitly when record was created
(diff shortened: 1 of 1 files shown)


vlad-psh:
Both bugs were fixed by #630 

I rebased this PR, so there's only new specs remained here (we had few use-cases missing in our specs)

danielmorrison (approved):

PR #630, opened 2022-07-05
Filter encrypted attributes automatically

vlad-psh:
Rails 7 introduces built-in attributes encryption: https://edgeguides.rubyonrails.org/active_record_encryption.html
Those are configured with `encrypts` keyword, eg: `encrypts :password`

Because these attributes are considered sensitive, we shouldn't write plain unencrypted values to our audits table by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,17 @@ User.auditing_enabled = false
 end
 ```
 
+### Encrypted attributes
+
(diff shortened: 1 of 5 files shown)


danielmorrison (changes requested):
This seems smart. Could you add a note to the README about this too?

vlad-psh:
@danielmorrison Thanks for comment!
I added small note to the README and also fixed specs for Rails < 7

danielmorrison (approved):

dsinn (commented):
We don't use `config.audit_class` and we're still a long way from being able to use ActiveRecord encryption (Rails 7).

PR #648, opened 2023-01-04
Allow Audited.audit_class= to receive a string class name and return …

rocket-turtle:
…an class

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -13,12 +13,15 @@ module Audited
     attr_writer :audit_class
 
     def audit_class
-      @audit_class ||= "Audited::Audit"
+      # The audit_class is set as String in the initializer. It can not be constantized during initialization and must
+      # be constantized at runtime. See https://github.com/collectiveidea/audited/issues/608
+      @audit_class = @audit_class.constantize if @audit_class.is_a?(String)
+      @audit_class ||= Audited::Audit
     end
 
-    def audit_model
-      audit_class.constantize
-    end
+    # remove audit_model in next major version it was only shortly present in 5.1.0
+    alias_method :audit_model, :audit_class
+    deprecate audit_model: "use Audited.audit_class instead of Audited.audit_model. This method will be removed."
 
     def store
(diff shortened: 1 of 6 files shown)


rocket-turtle:
Would it be better to remove the String case and add that to the Readme

https://github.com/collectiveidea/audited/pull/609#issuecomment-1180495205

```
ActiveSupport.on_load :active_record do
  Audited.audit_class = CustomAudit
end
```

I do not understand why this should not work. maybe [phortx](https://github.com/phortx) can describe the problem.
https://github.com/collectiveidea/audited/pull/609#issuecomment-1214161106

lib/audited.rb
```diff
@@ -13,12 +13,13 @@ class << self
     attr_writer :audit_class
```
ekohl, line 13: Perhaps you should replace `attr_writer` with a manual implementation to avoid the overhead in the getter:
```ruby
def audit_class=(cls)
  @audit_class = cls.is_a?(String) ? cls.constantize : cls
end
```

rocket-turtle: I had it like this in my first draft but I think it would not help.
The constantize would be called directly on the setting (in the initializer).

I think it would be the same as writing: `Audited.audit_class = "CustomAudit".constantize`

ekohl: Oh, that's an interesting thing. Perhaps a comment in the `audit_class` why it's there (like "it has to be lazy") could help future readers.

lib/audited.rb
```diff
@@ -13,12 +13,15 @@ class << self
...
 
     def audit_class
-      @audit_class ||= "Audited::Audit"
+      # The audit_class is set as String in the initializer. It can not be constantize during initialization and must be
```
ekohl, line 16:
```suggestion
      # The audit_class is set as String in the initializer. It can not be constantized during initialization and must be
```

ekohl:
Thanks for this. I've opened https://github.com/theforeman/foreman/pull/9595 to unpin again. Let's see if CI agrees this works :)

PR #657, opened 2023-02-13
Audit touch calls

mcyoung:
Touch is a form of `update` which generally doesn't call callbacks.  It does, however, invoke an `after_touch` callback that can be leveraged to store away audit information.  This can be informative with fields such as `locked_at` housing a timestamp for when a user account is locked and `nil` when the user's account is unlocked.  Having historical information about when a user was locked/unlocked and auditing comments around the why, is handy.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -231,23 +231,13 @@ module Audited
 
       private
 
-      def audited_changes
-        all_changes = respond_to?(:changes_to_save) ? changes_to_save : changes
-        filtered_changes = \
-          if audited_options[:only].present?
-            all_changes.slice(*self.class.audited_columns)
-          else
-            all_changes.except(*self.class.non_audited_columns)
-          end
+      def audited_changes(for_touch: false)
+        all_changes = if for_touch
+                        previous_changes
+                      else
+                        respond_to?(:changes_to_save) ? changes_to_save : changes
+                      end
 
-        filtered_changes = redact_values(filtered_changes)
-        filtered_changes = filter_encrypted_attrs(filtered_changes)
-        filtered_changes = normalize_enum_changes(filtered_changes)
-        filtered_changes.to_hash
-      end
-
-      def touch_changes
-        all_changes = previous_changes
         filtered_changes = \
           if audited_options[:only].present?
             all_changes.slice(*self.class.audited_columns)
@@ -341,7 +331,7 @@ module Audited
       end
 
       def audit_touch
-        unless (changes = touch_changes).empty?
+        unless (changes = audited_changes(for_touch: true)).empty?
           write_audit(action: "update", audited_changes: changes,
             comment: audit_comment)
         end
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -84,6 +84,7 @@ module Audited
 
         after_create :audit_create if audited_options[:on].include?(:create)
         before_update :audit_update if audited_options[:on].include?(:update)
+        after_touch :audit_touch if audited_options[:on].include?(:touch)
         before_destroy :audit_destroy if audited_options[:on].include?(:destroy)
 
         # Define and set after_audit and around_audit callbacks. This might be useful if you want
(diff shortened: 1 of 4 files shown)


lib/audited/auditor.rb
```diff
@@ -324,6 +330,13 @@ def audit_update
         end
       end
 
+      def audit_touch
+        unless (changes = audited_changes(for_touch: true)).empty?
+          write_audit(action: "update", audited_changes: changes,
```
mcyoung, line 335: The `action` could be set to `"touch"`, but I left it as an `"update"` since the underlying change is effectively updating a timestamp from 1 value to the current time.

danielmorrison: I think you made the right call.

spec/audited/auditor_spec.rb
```diff
@@ -414,6 +414,45 @@ def non_column_attr=(val)
     end
   end
 
+  describe "on touch" do
+    before do
+      @user = create_user(name: "Brandon", status: :active, audit_comment: "Touch")
+    end
+
+    around do |example|
+      freeze_time do
```
danielmorrison, line 423: Looks like freezing won't work on old Rails, and different DBs are storing it differently. I'd suggest checking that it is [within](https://relishapp.com/rspec/rspec-expectations/v/3-12/docs/built-in-matchers/be-within-matcher) a second of Time.current instead.

lib/audited/auditor.rb
```diff
@@ -230,8 +231,13 @@ def revision_with(attributes)
...
       private
 
-      def audited_changes
-        all_changes = respond_to?(:changes_to_save) ? changes_to_save : changes
+      def audited_changes(for_touch: false)
+        all_changes = if for_touch
+                        previous_changes
+                      else
+                        respond_to?(:changes_to_save) ? changes_to_save : changes
+                      end
```
danielmorrison, line 239: This seems a bit clearer to me, but 🤷 
```suggestion
        all_changes = if for_touch
                        previous_changes
                      elsif respond_to?(:changes_to_save)
                        changes_to_save
                      else
                        changes
                      end
```

danielmorrison (commented):
I like it. Let's get those tests passing.

spec/audited/auditor_spec.rb
mcyoung: Good suggestion, updated!

danielmorrison (approved):

mcyoung:
@danielmorrison Looking into the spec failures, it seems that Rails 5 -> 6 had a change under the hood w/ `touch` that impacted how `previous_changes` are reported.  Do you have feelings/thoughts on adjusting this to be a 6+ feature?

danielmorrison:
>  Do you have feelings/thoughts on adjusting this to be a 6+ feature?

That seems like a fine solution. We already have checks for `ActiveRecord::VERSION::MAJOR >= 6`.

jasonperrone:
How do I opt out of this on a given model? I have a model with an associated audit like so:
```
class RolesUsers < ApplicationRecord
 
  audited associated_with: :user, on: [:create, :destroy]
  
  belongs_to :role
  belongs_to :user, touch: true

end
```
When creating a RolesUsers record it touches :user as you can see. I don't want that to result in an audit record for user. How do I stop this change from happening?

The reason I want that is that my workflow creates a user instance and then add a RolesUsers to a collection on user and then saves user. This normally yields a create audit on user and a create audit on roles_users. But I am subsequently getting an additional update audit on the user I just created. I don't need the create AND the update audit in this situation.

mcyoung:
> How do I opt out of this on a given model? I have a model with an associated audit like so:
> 
> ```
> class RolesUsers < ApplicationRecord
>  
>   audited associated_with: :user, on: [:create, :destroy]
>   
>   belongs_to :role
>   belongs_to :user, touch: true
> 
> end
> ```
> 
> When creating a RolesUsers record it touches :user as you can see. I don't want that to result in an audit record for user. How do I stop this change from happening?
> 
> The reason I want that is that my workflow creates a user instance and then add a RolesUsers to a collection on user and then saves user. This normally yields a create audit on user and a create audit on roles_users. But I am subsequently getting an additional update audit on the user I just created. I don't need the create AND the update audit in this situation.

@jasonperrone Great call out, and I've noticed this as well and am working towards a fix.  For now, I think you should be able to update your audit call on the user to be `audited on: %i[create update destroy]` to exclude the new touch audit call.

jasonperrone:
Makes sense, thanks @mcyoung

mcyoung:
https://github.com/collectiveidea/audited/pull/660 should help to resolve double auditing for touches.

jasonperrone:
Cool @mcyoung , thank you.

PR #660, opened 2023-02-19
Don't double audit on touch for create / update actions

mcyoung:
When introducing the touch audit, we overlooked touch on association (`belongs_to :model, touch: true`) and this PR resolves that.  This comes up when a model has an association setup w/ a touch, which could possibly be behind the scenes like when it comes to ActiveStorage objects.

E.g. W/ a basic model setup
```ruby
module Company < ApplicationRecord
  has_many :employees
end

module Employee < ApplicationRecord
  belongs_to :company, touch: true
end
```
If a company is created w/ an employee included...
```ruby
company = Company.create!(name: "Company A", employees_attributes: [{ name: "Employee 1", hired_at: Time.zone.current }])
```
Before this update we would end up with a "create" and an "update" audit for the company where the "update" audit is essentially the same data as the "create" audit.  This PR essentially scrubs out the attributes if they match the prior audit during a touch audit.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -247,6 +247,13 @@ module Audited
             all_changes.except(*self.class.non_audited_columns)
           end
 
+        if for_touch
+          filtered_changes.reject! do |k, v|
+            audits.last.audited_changes[k].to_json == v.to_json ||
+            audits.last.audited_changes[k].to_json == v[1].to_json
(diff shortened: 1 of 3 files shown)


mcyoung:
@danielmorrison Just wanted to alert you to this, as it might be worth getting in as a minor update 🤷🏼

PR #662, opened 2023-02-21
Touch Audit - Ensure previous auditing

mcyoung:
If records are looking to audit for touch, but have yet to be audited, or are specifically auditing _just_ touch events, ensure we're not failing due to a bad previous audit check.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -249,6 +249,8 @@ module Audited
 
         if for_touch
           filtered_changes.reject! do |k, v|
+            next unless audits.present?
+
             audits.last.audited_changes[k].to_json == v.to_json ||
             audits.last.audited_changes[k].to_json == v[1].to_json
           end
--- a/spec/audited/auditor_spec.rb
+++ b/spec/audited/auditor_spec.rb
@@ -1,8 +1,8 @@
 require "spec_helper"
 
 # not testing proxy_respond_to? hack / 2 methods / deprecation of `version`
-# also, an additional 5 around `after_touch` for Versions before 6.
-uncovered = ActiveRecord::VERSION::MAJOR < 6 ? 14 : 9
+# also, an additional 6 around `after_touch` for Versions before 6.
+uncovered = ActiveRecord::VERSION::MAJOR < 6 ? 15 : 9
 SingleCov.covered! uncovered: uncovered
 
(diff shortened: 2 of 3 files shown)


mcyoung:
@danielmorrison Who woulda known touch events were such touchy things...

lib/audited/auditor.rb
```diff
@@ -249,6 +249,8 @@ def audited_changes(for_touch: false)
 
         if for_touch
           filtered_changes.reject! do |k, v|
+            next unless audits.present?
```
akostadinov, line 252: Shouldn't this be 
```rb
if for_touch && audits.present?
  ...
end
```

What is the point in iterating over the changes?

lib/audited/auditor.rb
```diff
@@ -249,6 +249,8 @@ def audited_changes(for_touch: false)
...
+
             audits.last.audited_changes[k].to_json == v.to_json ||
             audits.last.audited_changes[k].to_json == v[1].to_json
```
akostadinov, line 255: also it makes sense here to get `audits.last.audited_changes` once instead of twice per iteration.

akostadinov:
I would actually suggest to enable touch callback only if requested. It would usually not make sense to have it. Also it adds database queries to the audit which will add even more latency to responses.

lib/audited/auditor.rb
mcyoung: Great point, updated in https://github.com/collectiveidea/audited/pull/668

JDrizzy:
> Also it adds database queries to the audit which will add even more latency to responses

This scenario actually bit me when pushing the upgrade to production. Locally, the audits table only had a thousand records whereas production has millions 💥

If we could disable this callback in the config that was be awesome, as this would avoid the same scenario occurring if touch was forgotten from the options callback list. I can add this change if others agree it's needed? I was thinking of adding something that extends to all callbacks as well as it _might_ be beneficial for others. 

For example:
```Ruby
# audited.rb

Audited.config do |config|
  config.ignored_callbacks = [:create, :touch]
end

#####

# auditor.rb

after_create :audit_create if audit_callback?(:create)
...
after_touch :audit_touch if ::ActiveRecord::VERSION::MAJOR >= 6 && audit_callback?(:touch)

def audit_callback?(callback)
  return false if Audited.ignored_callbacks.include?(callback)

  audited_options[:on].include?(callback)
end
```

PR #664, opened 2023-02-26
Reflect Rails' serializer changes

nikz:
As a result of [Rails PR#47463](https://github.com/rails/rails/pull/47463), serialization is now type-checked, and so the approach of dynamically deciding when calling `#load` and `#dump` no longer works.

This means that you'll get the following errors/deprecation warnings when using Edge Rails with Audited:

```
DEPRECATION WARNING:                 Passing the class as positional argument is deprecated and will be remove in Rails 7.2.

                Please pass the class as a keyword argument:

                  serialize :audited_changes, type: Audited::YAMLIfTextColumnType
 (called from <class:Audit> at /Users/nik/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/audited-5.3.2/lib/audited/audit.rb:52)
```

And:

```
ActiveRecord::SerializationTypeMismatch: can't serialize `audited_changes`: was supposed to be a Audited::YAMLIfTextColumnType, but was a Hash. -- { ... }
```

(The first is just a syntax thing to be fair)

I don't know the best way to fix it to be honest - determining the column type would require connecting to the database on load, so that's sub-optimal. It might be possible to implement a `Coder` class that dynamically switches and passes the type checking?

This is the commit I'm using, which adds a configuration setting to choose json encoding by default (replaces the existing approach). This seems like something worth considering as it seems Rails has moved away from YAML for good reasons too. 

I've sent the PR in case it's useful, but of course it's a breaking change and I think there might be a better approach - figured I'd start the discussion anyway 😄 

Thank you!

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -10,7 +10,10 @@ module Audited
       :ignored_attributes,
       :max_audits,
       :store_synthesized_enums
-    attr_writer :audit_class
+    attr_writer \
+      :audit_class,
+      :encoding_type
+
 
     def audit_class
       # The audit_class is set as String in the initializer. It can not be constantized during initialization and must
@@ -22,6 +25,20 @@ module Audited
     # remove audit_model in next major version it was only shortly present in 5.1.0
     alias_method :audit_model, :audit_class
     deprecate audit_model: "use Audited.audit_class instead of Audited.audit_model. This method will be removed."
+
+    def encoding_type
+      @encoding_type ||= :json
+    end
+
+    def coder
+      if encoding_type == :json
+        ActiveRecord::Coders::JSON
+      elsif encoding_type == :yaml
+        ActiveRecord::Coders::YAMLColumn
+      else
+        ActiveRecord::Coders::ColumnSerializer
+      end
+    end
 
     def store
       current_store_value = Thread.current.thread_variable_get(:audited_store)
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -15,30 +15,6 @@ module Audited
   # * <tt>created_at</tt>: Time that the change was performed
   #
 
-  class YAMLIfTextColumnType
-    class << self
-      def load(obj)
-        if text_column?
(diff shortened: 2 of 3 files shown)


gjtorikian:
This is a significant upcoming change from Rails. 

@danielmorrison @tiagocassio Any chance this could be merged in? (Sorry for the random ping, I just grabbed the two most recent PR mergers I saw; since this has been open since February I'm not sure how actively this repo is watched). 

* * *

If this is PR indeed a breaking change, there's also another Rails deprecation notice nearby:

```
DEPRECATION WARNING: Module.deprecate without a deprecator is deprecated (called from singleton class at /Users/gjtorikian/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/audited-5.3.3/lib/audited.rb:25)
```

The code here suggests this can just be removed, too:

https://github.com/collectiveidea/audited/blob/379aeed7c22a530da0effa2b6b790ad806c5b7a1/lib/audited.rb#L23-L25

gjtorikian:
No worries—happy to help clean up / merge the present PRs, or do anything else that would be contributive. 👍

gjtorikian:
👋 @danielmorrison Rails 7.1 is coming out soon, so the deprecation here will be a failure in Rails 7.2. 

Note that the PR here won't get rid of all the deprecation warnings; there's [another line](https://github.com/collectiveidea/audited/pull/664#issuecomment-1529736313) I pointed to earlier. I don't think there's a PR open for that removal but happy to open one up.

gjtorikian:
BTW, I came across this block of code which would achieve the same future forward change without breaking any semver compatibility: https://github.com/gauravtiwari/noticed/blob/d14878a95602e56967f3b558b402f7f863608d91/lib/noticed/model.rb#L16C1-L19

gjtorikian:
Looks like the newer https://github.com/collectiveidea/audited/pull/686 resolved this.

danielmorrison:
Fixed via #686

PR #668, opened 2023-03-16
Touch Audit - a little cleanup

mcyoung:
As pointed out by @akostadinov in the last PR, [here](https://github.com/collectiveidea/audited/pull/662#discussion_r1124546706) this PR cleans up the touch looking at previous audit logic a bit.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -247,12 +247,11 @@ module Audited
             all_changes.except(*self.class.non_audited_columns)
           end
 
-        if for_touch
+        if for_touch && audits.present?
+          last_audit = audits.last.audited_changes
           filtered_changes.reject! do |k, v|
-            next unless audits.present?
-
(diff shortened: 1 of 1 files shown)


mcyoung:
@danielmorrison just a bump if you feel this is worthwhile, thanks!

lib/audited/auditor.rb
```diff
@@ -247,12 +247,11 @@ def audited_changes(for_touch: false)
             all_changes.except(*self.class.non_audited_columns)
           end
 
-        if for_touch
+        if for_touch && audits.present?
+          last_audit = audits.last.audited_changes
```
akostadinov, line 251: Another small optimization. No need to perform 2 queries - for the `#present?` and to get the last one. If there are no audits, then the query for the last one would result in a `nil`.

```suggestion
        if for_touch && (last_audit = audits.last)
          changes = last_audit.audited_changes
```

Issue #601, opened 2021-11-01
Incorrect User in Audit


sauloarruda:
Hello,

We are using the audit to generate around 20 thousand records daily. In some cases we find that the user who performed the action is incorrect. Analyzing the code we saw that the Thread.current[:audited_store] method is called to get the current user. We use AWS with loadbalancer with 2 to 4 servers at peak hours.

We need help to understand why this is happening.

tiagocassio:
@sauloarruda did you solve this problem? I'm getting the same problem here, im some cases the user that makes the change is not the correct one.

sauloarruda:
Hi @tiagocassio 
We solved using RequestStore. This is the monkey patch. I hope it helps you too.

```module Audited
  class << self
    def store
      current_store_value = RequestStore.store[:audited_store]

      if current_store_value.nil?
        RequestStore.store[:audited_store] = {}
      else
        current_store_value
      end
    end
  end

  # add sudo_user_id
  class Sweeper
    STORED_DATA = {
      current_remote_address: :remote_ip,
      current_request_uuid: :request_uuid,
      current_user: :current_user,
      sudo_user_id: :sudo_user_id
    }

    def sudo_user_id
      controller.try(:sudo_user_id)
    end
  end

  class Audit
    def set_audit_user
      self.user ||= ::Audited.store[:audited_user] # from .as_user
      self.user ||= ::Audited.store[:current_user].try!(:call) # from Sweeper
      self.sudo_user_id ||= ::Audited.store[:sudo_user_id]
      nil # prevent stopping callback chains
    end
  end
end
```

sauloarruda:
the sudo user part is a new implementation, ignore it please.

tiagocassio:
Hi @sauloarruda! I've already proposed a PR implementing RequestStore. See https://github.com/collectiveidea/audited/pull/669

the-spectator:
@sauloarruda @tiagocassio I have created https://github.com/collectiveidea/audited/pull/673 as an alternative attempt to solve the issue using `ActiveSupport::CurrentAttributes`. If possible can you test your use case against my branch?

`gem 'audited', git: 'https://github.com/the-spectator/audited', branch: 'migrate_to_current_attributes'`

jlledom:
[...]

PR #669, opened 2023-04-19
Use RequestStore instead of Thread.current for thread-safe requests.

tiagocassio:
Related to: https://github.com/collectiveidea/audited/issues/601 and inspired by: https://github.com/paper-trail-gem/paper_trail/commit/e28d29ef0f58cded1537df57551b73ac5b081b36

### DIFF
--- a/audited.gemspec
+++ b/audited.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = ">= 2.3.0"
 
   gem.add_dependency "activerecord", ">= 5.0", "< 7.1"
+  gem.add_dependency "request_store", "~> 1.2"
 
   gem.add_development_dependency "appraisal"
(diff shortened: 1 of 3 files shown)


danielmorrison:
This is good. Thanks!

danielmorrison (approved):

danielmorrison (changes requested):
Looks like the tests need some work.

danielmorrison (approved):

mroach:
What's the motivation with using another gem as opposed to [ActiveSupport::CurrentAttributes](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html)?  My understanding of `CurrentAttributes` is that it's designed specifically to deal with this problem, and it's already available due to the indirect dependency via `ActiveRecord`.

danielmorrison:
I like the idea of moving to ActiveSupport::CurrentAttributes. I'd love to see a PR with that.

PR #686, opened 2023-09-24
Add support for Rails 7.1

yuki24:
This PR adds:

1. A new gemfile and build matrix for Rails 7.1
2. Fixes to address deprecation warnings
3. Test updates to make it compatible with the latest Psych

I will leave comments to each change to provide more context. Addresses #685

PR #689, opened 2023-11-09
Robustify Rails version checks

blaet:
Ensure Rails versions are properly checked, instead of relying on `String` comparisons.

Example of why this is an issue
```ruby
[8] pry(main)> Rails.version > "10.2.3"
=> true
[9] pry(main)> Gem::Version.new(Rails.version) >= Gem::Version.new("10.2.3")
=> false
```
This example was taken from [this comment](https://github.com/collectiveidea/audited/pull/686#discussion_r1353239732) on a recent PR.

## Approach

This change leverages the already existing function `Rails.gem_version`. This has been part of the Rails codebase for 10+ years, making it compatible with all supported Rails versions.

### DIFF
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -49,7 +49,7 @@ module Audited
     cattr_accessor :audited_class_names
     self.audited_class_names = Set.new
 
-    if Rails.version >= "7.1"
+    if Rails.gem_version >= Gem::Version.new("7.1")
       serialize :audited_changes, coder: YAMLIfTextColumnType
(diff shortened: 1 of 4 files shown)


akostadinov (approved):
Thank you!

blaet:
Fixed a small syntax error. Build should now be 🍏

blaet:
Hi @danielmorrison, any chance you could have a look at this one? Thanks so much! 🚀

danielmorrison (approved):

PR #694, opened 2024-01-15
 New configurable settings: encryption for audited_changes & filtering encrypted attrs

### LINKED ISSUE #690: Encryption for audited_changes / disabling the FILTERED feature
We recently set up Activerecord 7 encryption in our app, which was already using Audited, and then realized we needed to disable the automatic filtering of AR-encrypted fields for the sake of our record keeping policy. But the data still needed to be stored encrypted (also a matter of policy). So, we encrypted the entire audited_changes column for the Audit model. This required a local patch.

Setting up AR to encrypt the whole field was pretty straightforward. This module override goes in a library loaded in our initializers:
```
module Audited
  class Audit < ::ActiveRecord::Base
    encrypts :audited_changes
  end
end
```
Note that you would also need to ensure you have
`config.active_record.encryption.support_unencrypted_data = true`
in your application config so AR doesn't choke on older records. (You can avoid needing this by deleting or retroactively encrypting the existing records.)

We then overrode the specific classes that had encrypted fields, with another module:
```
module UnfilteredAudit
  def filter_encrypted_attrs(input)
    input # i.e. actually DON'T filter the encrypted attrs since we are encrypting the column altogether
  end
end
```
and then adding `include UnfilteredAudit` after the `audited` line in the specific models. 

Both these changes are a bit fragile and slightly hacky. So my issue is really a request to make the filtering optional/configurable and to add an option to use encryption on the whole column, both of which could be handled with a couple simple if statements in audited. Note, of course, once you encrypt the audited_changes column you will not be able to directly query the table data in SQL, but it can still be searched programmatically post-decryption. 

I would add a PR for this if anyone is interested, but it would likely be a long wait (and maybe these suggestions bear further discussion).

sriddbs:
Closes #690

* Add new configurable vars:
  * `encrypt_audited_changes` default to `false`
  * `filter_encrypted_attributes` default to `true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,17 @@ class User < ActiveRecord::Base
 end
 ```
 
+You can disable the filtering by adding a config
+
+```ruby
+Audited.filter_encrypted_attributes = false
+```
+
+If you want to encrypt the changes that are audited, you can simply add this line to your config
+```ruby
+Audited.encrypt_audited_changes = true
+```
+
 ### Custom `Audit` model
 
 If you want to extend or modify the audit model, create a new class that
--- a/lib/audited/audit.rb
+++ b/lib/audited/audit.rb
@@ -55,6 +55,10 @@ module Audited
       serialize :audited_changes, YAMLIfTextColumnType
     end
 
+    if Rails.gem_version >= Gem::Version.new("7.0") && Audited.encrypt_audited_changes
+      encrypts :audited_changes
+    end
+
     scope :ascending, -> { reorder(version: :asc) }
     scope :descending, -> { reorder(version: :desc) }
     scope :creates, -> { where(action: "create") }
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -8,6 +8,8 @@ module Audited
     attr_accessor \
       :auditing_enabled,
       :current_user_method,
+      :encrypt_audited_changes,
+      :filter_encrypted_attributes,
       :ignored_attributes,
       :ignored_default_callbacks,
       :max_audits,
@@ -40,6 +42,8 @@ module Audited
 
   @current_user_method = :current_user
   @auditing_enabled = true
+  @encrypt_audited_changes = false
+  @filter_encrypted_attributes = true
   @store_synthesized_enums = false
 end
 
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -289,6 +289,8 @@ module Audited
       end
 
       def filter_encrypted_attrs(filtered_changes)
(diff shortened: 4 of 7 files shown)


sriddbs:
@danielmorrison please review and let me if the proposed config in the PR is good, also I wonder why some of the specs are failing 🤔

README.md
```diff
@@ -403,6 +403,17 @@ class User < ActiveRecord::Base
 end
 ```
 
+You can disable the filtering by adding a config
+
+```ruby
+Audited.filter_encrypted_attributes = false
+```
+
+If you want to encrypt the changes that are audited, you can simply add this line to your config
```
danielmorrison, line 412: Maybe another header here to show that these are separate features?

danielmorrison (commented):
Looking good to me. Test errors are something strange with coverage? Doesn't make sense offhand.

sriddbs:
> Looking good to me. Test errors are something strange with coverage? Doesn't make sense offhand.

I had the same errors with `spec/audited/audit_spec`,  rspec clearly showed me the errors locally and I could fix it

for this it just runs fine

```
rspec spec/audited/auditor_spec.rb
Created database ':memory:'
............................................................................................................*.....*........................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Audited::Auditor without auditing should be thread safe using a #without_auditing block
     # No reason given
     # ./spec/audited/auditor_spec.rb:996

  2) Audited::Auditor with auditing should be thread safe using a #with_auditing block
     # No reason given
     # ./spec/audited/auditor_spec.rb:1077
```

sriddbs:
@danielmorrison if this PR looks good, please merge?

lib/audited/audit.rb
```diff
@@ -55,6 +55,10 @@ class Audit < ::ActiveRecord::Base
       serialize :audited_changes, YAMLIfTextColumnType
     end
 
+    if Rails.gem_version >= Gem::Version.new("7.0") && Audited.encrypt_audited_changes
```
tvararu, line 58: This doesn't seem to work as expected, it appears that this is evaluated before the `Audited.encrypt_audited_changes = true` in the initializer runs so it evaluates to false initially and is never re-evaluated. The `encrypts` call below never runs as a result.

sb-truefalse:
How are you doing with PR?

lib/audited/audit.rb
```diff
@@ -55,6 +55,10 @@ class Audit < ::ActiveRecord::Base
...
+      encrypts :audited_changes
```
sb-truefalse, line 59: Is it necessary to take into account the deterministic encryption?

PR #673, opened 2023-05-01
Replace RequestStore with `ActiveSupport::CurrentAttributes`

the-spectator:
Followup of https://github.com/collectiveidea/audited/pull/669

This PR is an attempt to use `ActiveSupport::CurrentAttributes` instead of the `RequestStore` gem.

PR #702, opened 2024-04-02
Replace RequestStore with ActiveSupport::CurrentAttributes

punkisdead:
Based on the work done here: https://github.com/collectiveidea/audited/pull/673 by @the-spectator

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,6 @@ jobs:
       matrix:
         ruby: [2.3, 2.4, 2.5, 2.6, 2.7, 3.0, 3.1, 3.2]
         appraisal:
-          - rails50
-          - rails51
           - rails52
           - rails60
           - rails61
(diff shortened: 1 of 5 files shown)


punkisdead:
Since `ActiveSupport::CurrentAttributes` was introduced in Rails 5.2 [Rails 5.2 Release](https://guides.rubyonrails.org/5_2_release_notes.html#active-support-notable-changes) I had to increase the minimum required version of `activesupport` and `activerecord` to 5.2.

I'm not entirely sure why the Fiber spec would pass on Rails 7.0 and not previous versions, but I'm not entirely sure we need that spec in this gem? ¯\_(ツ)_/¯ 

Even in current Rails, the default isolation level is set to `:thread` according to [Configuring Rails](https://guides.rubyonrails.org/v7.0/configuring.html#config-active-support-isolation-level)

punkisdead:
I can take a look at it...

danielmorrison:
I can't remember offhand why we added the fiber spec. Probably sanity check for… something. 🤷

danielmorrison (approved):

punkisdead:
giphy

PR #715, opened 2024-06-13
Check if audits is loaded before executing database queries

wkirby:
This should be a no-op change that allows for performance improvements by preloading the `audits` association. By checking if the `audits` association is already loaded, we can avoid N+1 queries by operating on the already loaded audits in ruby instead of going to the database again to check if a version exists, etc.

This code should be _functionally_ identical to the existing code, but allows for something like:

```ruby
@posts = Post.all.include(:audits)

@posts_at_v1 = @posts.map do |post|
  post.revision(1)
end
```

To avoid repeated database access.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -143,9 +143,9 @@ module Audited
       #   end
       #
       def revisions(from_version = 1)
-        return [] unless audits.from_version(from_version).exists?
+        return [] unless has_version?(from_version)
 
-        all_audits = audits.select([:audited_changes, :version, :action]).to_a
+        all_audits = audits.pluck(:audited_changes, :version, :action).to_a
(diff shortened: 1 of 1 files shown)


danielmorrison:
Tests are failing, but I like the idea.

wkirby:
@danielmorrison thanks, will fix test failures

wkirby:
@danielmorrison specs are now passing _except_ for code coverage (see the test run on our fork here: https://github.com/apsislabs/audited/actions/runs/10538941999?pr=1). I believe this is because of the addition of the new `loaded?` branches being untested.

I'm happy to either:

* Update coverage exceptions so the tests pass
* Attempt to add some specs to cover these branches

wkirby:
@danielmorrison just thought I'd poke in here and see if you had an answer for how you wanted me to handle the new untested lines?

PR #717, opened 2024-06-21
Add support to rails 8.0

fernandomenolli:
This PR will:

- Add support to rails 8.0


danielmorrison (approved):

Appraisals
```diff
@@ -55,3 +55,10 @@ appraise "rails71" do
   gem "pg", ">= 1.1"
   gem "sqlite3", ">= 1.4"
 end
+
+appraise "rails80" do
+  gem "rails", ">= 7.1.0.beta1", "< 8.1"
```
chaadow, line 60: @danielmorrison @fernandomenolli 

This is not rails 8. Rails 8 is the `main` branch which is `8.0.0.alpha` which is **NOT** released on rubygems.

this will target at best `7.2.x`

.github/workflows/ci.yml
```diff
@@ -78,6 +79,8 @@ jobs:
             ruby: 2.5
           - appraisal: rails71
             ruby: 2.6
+          - appraisal: rails80
+            ruby: 2.6
```
chaadow, line 83: rails 8 depends on ruby 3.1 minimum

audited.gemspec
```diff
@@ -16,11 +16,11 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.3.0"
 
-  gem.add_dependency "activerecord", ">= 5.2", "< 7.2"
-  gem.add_dependency "activesupport", ">= 5.2", "< 7.2"
+  gem.add_dependency "activerecord", ">= 5.2", "< 8.1"
```
chaadow, line 19: @danielmorrison setting `< 8.1` is the same as supporting rails `main` branch, which is "edge rails"

So why not just removing the upper bound like so `gem.add_dependency "activerecord", ">= 5.2"` so that PRs like this won't be necessary, and future PRs by contributors will be about fixing bugs on newer versions of rails, instead of these "Bump rails " kind of PRs

PR #734, opened 2024-10-21
allow calling audited multiple times

### LINKED ISSUE #731: Make `audited` class method idempotent
This is an issue that's caused major headaches for me in many of my projects, because `audited` can only be called once. Calling it a second time does not function as intended, and worse, it doesn't error out if it's been called prior.

The use case is simple--I want all my models to be audited unless I opt out of the process. Whether its my careless mistake or a team member that's unaware of it, it's quite easy to miss calling `audited` when you create a new model. To solve this problem, I've created an `ApplicationRecord` class that calls `audited if: :audit?`, from which all other classes inherit, the default class' `audit?` method returns `true`, and I can opt out on a per-model basis. Cool.

## Problem 1

Options can't be amended for models that need to use them. If I want to use `only:`/`except:`/`associated_with:`, these options are totally ignored on the second invocation. So now I have several years' worth of audit history all clogged up with ephemeral attributes that I didn't want to trigger an audit record.

## Problem 2

`self.primary_key = :name` is only respected if called _before_ `audited` is called. So now I have several years' worth of audit history with NULL `auditable_id` attributes, and have no idea what records they actually belong to.

## Solution

Make the `audited` call idempotent so we can call it more than once. The options from the latest invocation should be respected, or at the very least, trigger an error to warn of unintended consequences. Many of the calls in the `audited` class method are already idempotent, but I would like the options to be respected or merged so we can call this on an abstract model, and call it again on the concrete model if needed.

mohammednasser-32:
Allow `audited` to be called multiple times

When `audited` is called for the first time, setup audit normally

However if `audited` is called on a class that is already audited we update `audit_options` and log a warning message

an example of the warning message logged:

Fixes #731

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -59,9 +59,16 @@ module Audited
       #     end
       #
       def audited(options = {})
-        # don't allow multiple calls
-        return if included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
+        audited? ? update_audited_options(options) : set_audit(options)
+      end
 
+      private
+
+      def audited?
+        included_modules.include?(Audited::Auditor::AuditedInstanceMethods)
+      end
+
+      def set_audit(options)
         extend Audited::Auditor::AuditedClassMethods
         include Audited::Auditor::AuditedInstanceMethods
 
@@ -69,10 +76,7 @@ module Audited
         class_attribute :audited_options, instance_writer: false
         attr_accessor :audit_version, :audit_comment
 
-        self.audited_options = options
-        normalize_audited_options
-
-        self.audit_associated_with = audited_options[:associated_with]
+        set_audited_options(options)
 
         if audited_options[:comment_required]
           validate :presence_of_audit_comment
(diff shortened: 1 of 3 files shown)


danielmorrison (approved):

arcreative:
@mohammednasser-32 awesome, thanks for the quick turnaround! Just curious—if the updated options work the same as if it were originally called once with those options, maybe the warning is unnecessary? My suggestion about the warning was in lieu of a fix so people wouldn’t hit the same pitfall as I did and not know it :-)

mohammednasser-32:
@arcreative so would it be better to remove the warning?

arcreative:
I think so. My original suggestion for the warning was to tell people that it wasn't working as intended. If it's working as intended, then I think the warning is unnecessary.

mohammednasser-32:
@arcreative perfect, incorporated [here](https://github.com/collectiveidea/audited/pull/734/commits/882d1cb7742b7c779c5f6612578353e11bfa9f54)

mohammednasser-32:
Hey @danielmorrison , since this already got approved I was just wondering if it is going to be merged soon? Thank you 🙏

danielmorrison (approved):

PR #735, opened 2024-10-29
Introduce support for audited_class setting per model

amkisko:
Done:
- test coverage for all changes in this PR
- README updated accordingly
- generators updated to support custom table name

Background:
- single table might become problematic for large amount of logs/audits
- partitioning of audits table leads to other problems to solve
- having full separation of audits tables per model gives more flexibility
- keeping it strict and as much verbose as needed
- compatibility with previous behavior required
- looked at https://github.com/collectiveidea/audited/pull/641

PR #736, opened 2024-10-29
Additional context

amkisko:
Done:
- updated README with additional instructions
- implemented `context` storage and merging before audit record writing
- implemented test coverage for changed parts
- updated generators and templates

Background:
- `comment:string` is not enough for storing and accessing additional context
- known approach is to have separate column with additional serialized data
- it could be nice to have global setter to add context once
- it should be always optional to add more context

References:
- it will have minor conflicts with https://github.com/collectiveidea/audited/pull/735
- conflicts are resolved fully in https://github.com/amkisko/audited/tree/release

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,26 @@ You can ignore the default callbacks globally unless the callback action is spec
 Audited.ignored_default_callbacks = [:create, :update] # ignore callbacks create and update
 ```
 
+### Context
+
+You can attach context to each audit using an `audit_context` attribute on your model.
+
+```ruby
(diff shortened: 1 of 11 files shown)


amkisko:
This one requires improvement, I completely missed the point that `audited_changes` is stored as `text`. `context` should be same for better support.

amkisko:
@danielmorrison fixes pushed here.

amkisko:
Same fixes applied here from https://github.com/collectiveidea/audited/pull/735

amkisko:
@danielmorrison Hey! Merge conflict resolved here too.

PR #754, opened 2025-03-06
Optionally pass audited action to conditionals

sebastianludwig:
I only want to record destructions under _some_ circumstances. The proposed implementation allows users to optionally accept an `action` parameter in their condition method or Proc. I've implemented tests and updated the documentation. Looking forward to your feedback!

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -33,7 +33,7 @@ module Audited
       #       audited except: :password
       #     end
       #
-      # * +require_comment+ - Ensures that audit_comment is supplied before
+      # * +comment_required+ - Ensures that audit_comment is supplied before
       #   any create, update or destroy operation.
       # * +max_audits+ - Limits the number of stored audits.
 
@@ -375,7 +375,7 @@ module Audited
       def write_audit(attrs)
         self.audit_comment = nil
 
-        if auditing_enabled
+        if auditing_enabled(attrs[:action])
           attrs[:associated] = send(audit_associated_with) unless audit_associated_with.nil?
 
           run_callbacks(:audit) {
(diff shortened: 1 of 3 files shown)


lib/audited/auditor.rb
```diff
@@ -33,7 +33,7 @@ module ClassMethods
       #       audited except: :password
       #     end
       #
-      # * +require_comment+ - Ensures that audit_comment is supplied before
+      # * +comment_required+ - Ensures that audit_comment is supplied before
```
sebastianludwig, line 36: Kind of related documentation fix: The private method is called `require_comment`, the option is called `comment_required`.

lib/audited/auditor.rb
```diff
@@ -430,18 +438,21 @@ def require_comment
...
+          run_conditional_check(audited_options[:unless], action: action, matching: false) &&
           self.class.auditing_enabled
       end
 
-      def run_conditional_check(condition, matching: true)
+      def run_conditional_check(condition, action:, matching: true)
         return true if condition.blank?
-        return condition.call(self) == matching if condition.respond_to?(:call)
-        return send(condition) == matching if respond_to?(condition.to_sym, true)
+        result = case condition
```
sebastianludwig, line 449: Rewritten in the style of `evaluate_max_audits` above

lib/audited/auditor.rb
```diff
@@ -430,18 +438,21 @@ def require_comment
...
+        when Proc then condition.call(self, action)
```
sebastianludwig, line 450: I thought about passing `action` as Symbol but `Audit.action` is also a String so it's more consistent this way.

sebastianludwig:
It would be easy to also incorporate https://github.com/collectiveidea/audited/pull/697 into this PR, but I'm not sure it's worth it. Using `proc { MyApp::CONFIG[:audit_my_model] }` as workaround seems simple enough to avoid the complexity.

PR #755, opened 2025-03-10
Ensure that no audit-related queries run when auditing is disabled

rafaelsales:
# Problem

Calling `Models::ActiveRecord::User.without_auditing { user.touch }` generates the query below:

```sql
SELECT "audits".* FROM "audits" WHERE "audits"."auditable_id" = ? AND "audits"."auditable_type" = ? ORDER BY "audits"."version" DESC LIMIT ?
```

Culprit: https://github.com/collectiveidea/audited/blob/f935285c27151dff179ef274776f865302d40b8d/lib/audited/auditor.rb#L361-L362

### Why this is a big deal

In an endpoint that requires high performance, I need to perform `record.touch(:connected_at)` along with some other audited record updates and make sure we're not running any extra queries.

# Solution

The current code checks for `auditing_enabled` too deep in the callback chain. Instead, we should use guards to return in each callback as early as possible.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -79,7 +79,7 @@ module Audited
         set_audited_options(options)
 
         if audited_options[:comment_required]
-          validate :presence_of_audit_comment
+          validate :validate_presence_of_audit_comment
           before_destroy :require_comment if audited_options[:on].include?(:destroy)
         end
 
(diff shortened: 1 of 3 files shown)


connorshea:
This would be great to see merged, I just discovered it in our test suite at work.

PR #757, opened 2025-04-14
Allow enable/disable Audit for all classes but only locally

lucasarnaud:
It is already possible to disable Audits locally for a specific object or to an entire model

image

but it not possible to do this to all classes. The only way would be to disable it at the module level with `Audited.auditing_enabled = false`, but this will disable audits for the entire process, which is undesirable in a multi-threaded environment like Sidekiq or Puma.

So I am opening this PR to suggest/add this feature.

Let me know if makes sense for the gem or if i need to make any changes to make it accepted.

### DIFF
--- a/lib/audited.rb
+++ b/lib/audited.rb
@@ -10,13 +10,12 @@ module Audited
 
   class << self
     attr_accessor \
-      :auditing_enabled,
       :current_user_method,
       :ignored_attributes,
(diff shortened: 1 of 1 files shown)


lucasarnaud:
The proposed changes would work even if the `with` and `without` blocks are nested as in the example:

image

Issue #772, opened 2026-02-27
Audited.current_user_method fails silently if specified method not defined


arcreative:
Having some issues with the `Audited.current_user_method` configuration option falling back to to `null` user_id and user_type on audit entries when the method is not defined on the controller. https://github.com/collectiveidea/audited/blob/dbf8432604ec97734bcfb3dc6df61f4faa0d1b84/lib/audited/sweeper.rb#L23 seems to be the offending line of code--the `if` statement should be omitted entirely, or alternatively, another exception should be raised that makes it more verbose than just a `NoMethodError` if the specified method is inaccessible. 

This is a very concerning issue to encounter with a gem that has to do with risk and compliance, and I think this warrants a CVE being raised, as it affects the integrity/accountability that was intended by implementing this gem in the first place. Issues like this should cause a failure of some sort, and do so as early as possible.

PR #773, opened 2026-04-04
Warn when Audited.current_user_method references undefined method

55728:
## Problem

When \`Audited.current_user_method\` is set to a method that doesn't exist on the controller, audited silently falls back to \`nil\`, resulting in audit records with no user. This makes misconfiguration very difficult to detect, especially in compliance-critical applications (#772).

## Solution

Emit a warning via \`Kernel#warn\` when the configured method is not defined on the controller, while preserving the existing behavior of recording a nil user. This avoids breaking existing applications while making the misconfiguration visible in logs.

A future major version could consider raising an error instead.

Ref #772


abdullah-bin-shamlan (approved):

55728:
Closing this as stale. Happy to reopen if needed. Thanks!

PR #774, opened 2026-05-04
Eliminate `audits.last` query when filtering touch changes

reid-rigo:
This change is an optimization for the fix introduced in https://github.com/collectiveidea/audited/pull/660

Cache the last audit changes on the instance to remove the `audits.last` query when filtering changes in `audited_changes`.

### DIFF
--- a/lib/audited/auditor.rb
+++ b/lib/audited/auditor.rb
@@ -267,7 +267,7 @@ module Audited
 
         filtered_changes = normalize_enum_changes(filtered_changes)
 
-        if for_touch && (last_audit = audits.last&.audited_changes)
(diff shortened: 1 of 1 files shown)


akostadinov:
How is this supposed to work for multiple processes working simultaneously? Can a few tests be added to demonstrate the desired behavior when memoized changes are present/not present/outdated by another process?

reid-rigo:
@akostadinov, this change is an optimization for the fix in https://github.com/collectiveidea/audited/pull/660 (added to PR description)

In main the code assumes that `audits.last` is the audit created within the current thread of execution, which won't be true in concurrent scenarios. This change should fix that by saving "these are the changes I just saved" in memory.

